PR #7, opened 2019-05-01
Adding support for inl packaging for rocThrust 

saadrahim:
inl files are used in rocThrust and need to be packaged for this header only library.


saadrahim:
@pfultz2 Can you assist with accepting these changes? These are needed to successfully package rocThrust. Math libraries uses rocm-cmake extensively and will be making more contributions to this project soon.

PR #9, opened 2019-06-28
export target remove library

saadrahim:
This is untested code, just showing what I believe is needed to fix the header only library export target issue.

### DIFF
--- a/share/rocm/cmake/ROCMInstallTargets.cmake
+++ b/share/rocm/cmake/ROCMInstallTargets.cmake
@@ -93,7 +93,7 @@ ${NAME}(${ARGS})
 endfunction()
 
 function(rocm_export_targets)
-    set(options)
+    set(options HEADER_ONLY)
     set(oneValueArgs NAMESPACE EXPORT NAME COMPATIBILITY PREFIX)
     set(multiValueArgs TARGETS DEPENDS INCLUDE)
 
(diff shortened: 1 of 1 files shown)


pfultz2:
Why do these lines need to be conditioned? The `LIBRARIES` variable should still be defined for header-only libraries.

saadrahim:
> Why do these lines need to be conditioned? The LIBRARIES variable should still be defined for header-only libraries.

Those lines generate request to link against .so files in the library_name-config.cmake file. The header only libraries have no "library" file to link against.

pfultz2:
> Those lines generate request to link against .so files

Why? It should just have `hip::hipcub` there. And looking [here](https://github.com/ROCmSoftwarePlatform/hipCUB/blob/master/hipcub/CMakeLists.txt#L33) it is just a `INTERFACE` library so there is no build artifact.

> The header only libraries have no "library" file to link against.

There should still be a cmake target that is linked, as header-only still have usage requirements. And whether its header-only or not, is an implementation detail, the user should still link against `hip::hipcub`.

For header-only libraries, an `INTERFACE` library is created, which is what is done.

saadrahim:
> Those lines generate request to link against .so files
> 
> Why? It should just have hip::hipcub there. And looking here it is just a INTERFACE library so there is no build artifact.
> 
> The header only libraries have no "library" file to link against.
> 
> There should still be a cmake target that is linked, as header-only still have usage requirements. And whether its header-only or not, is an implementation detail, the user should still link against hip::hipcub.
> For header-only libraries, an INTERFACE library is created, which is what is done.

The rocm-cmake version that you did not recognize the git commit version for some strange reason didn't print these lines. I was trying to reproduce that behavior as I knew it worked. However, your explanation makes sense. I will get QA/devops to use the master branch to solve the issues.

saadrahim:
@pfultz2 Going back to this request, what would you say if I told you a library uses the classic way of handling cmake dependencies? When enforcing these targets for that library, it fails to build.

I am not sure what to do. I have manually built two version of the libraries in question for 2.5/2.6 by disabling the two lines in this PR. This is not sustainable in the long term.

PR #10, opened 2019-07-11
softlink directory

saadrahim:
I need to install the soft links in a custom directory in CMAKE_INSTALL_PREFIX. What do you propose is the best way to do this? This PR contains untested suggestion. The goal is for a library to install softlinks to files in /opt/rocm/$library/include* to /opt/rocm/include/$library. Now by default the softlinks go to /opt/rocm/include/
If there is another way to do it, that will be great.

### DIFF
--- a/share/rocm/cmake/ROCMInstallSymlinks.cmake
+++ b/share/rocm/cmake/ROCMInstallSymlinks.cmake
@@ -3,7 +3,7 @@
 ################################################################################
 
 
-function(rocm_install_symlink_subdir SUBDIR)
+function(rocm_install_symlink_subdir SUBDIR CUSTOM_DIRECTORY)
     # TODO: Check if SUBDIR is relative path
     # Copy instead of symlink on windows
     if(CMAKE_HOST_WIN32)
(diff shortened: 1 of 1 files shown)


saadrahim:
@pfultz2 I cannot add you as review but please take a look.

pfultz2:
> The goal is for a library to install softlinks to files in /opt/rocm/$library/include* to /opt/rocm/include/$library.

You need to symlink the files not the directory. 

You can do `rocm_install_targets(... PREFIX ${LIB})` and then at the end do `rocm_install_symlink_subdir(${LIB})` where `${LIB}` is your library name. The `rocm_install_symlink_subdir` must come at the very end.

This will install in `${CMAKE_INSTALL_PREFIX}/${LIB}` and then symlink everything into `${CMAKE_INSTALL_PREFIX}`.

saadrahim:
The directory structure of rocRAND is different from the standard paths used by rocm-cmake. We are using rocm_install_targets as suggested. To use rocm-cmake for this library without forcing user changes needs us to add the extra directory after CMAKE_INSTALL_PREFIX. We can do this with a post install script but doing it in rocm-cmake might be cleaner.

pfultz2:
> The directory structure of rocRAND is different from the standard paths used by rocm-cmake.

I dont understand how you are wanting to install rocrand, but it should follow the standard paths. The files for rocrand would be installed into `/opt/rocm/rocrand` and then you would symlink everything into `/opt/rocm`. In this case, the user sets `CMAKE_INSTALL_PREFIX=/opt/rocm` like always, and then you would call `rocm_install_targets(... PREFIX rocrand)` and then at the end `rocm_install_symlink_subdir(rocrand)` to install the symlinks.

PR #11, opened 2019-07-29
Fix issues with changes to versioning

aaksud:
1. Remove "dirty" description from git command
    This will be added back by 2.8
2. Remove the extra $ symbol in from of ENV so that
   the env variables are actually passed down when
   setup version is used

### DIFF
--- a/share/rocm/cmake/ROCMSetupVersion.cmake
+++ b/share/rocm/cmake/ROCMSetupVersion.cmake
@@ -61,10 +61,10 @@ endfunction()
 
 function(rocm_get_build_info OUTPUT DELIM)
     set(_info)
-    if(DEFINED $ENV{JOB_NAME})
+    if(DEFINED ENV{JOB_NAME})
         set(_info ${_info}${DELIM}$ENV{JOB_NAME})
     endif()
-    if(DEFINED $ENV{BUILD_ID})
+    if(DEFINED ENV{BUILD_ID})
         set(_info ${_info}${DELIM}$ENV{BUILD_ID})
     endif()
     rocm_set_parent(${OUTPUT} ${_info})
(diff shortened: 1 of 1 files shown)


share/rocm/cmake/ROCMSetupVersion.cmake
```diff
@@ -87,7 +87,7 @@ function(rocm_get_version OUTPUT_VERSION)
     rocm_get_build_info(BUILD_INFO -)
 
     if(GIT)
-        set(GIT_COMMAND ${GIT} describe --dirty --long --match [0-9]*)
```
saadrahim: You should leave a comment that we will addthe dirty flag again for 2.8.

saadrahim: I think two PRs is best as one fix and another changes the behavior. The dirty removal will probably be reverted.

saadrahim: Maybe a branch for rocm 2.7 is best for this fix. RC candidate branch is the best approach going forward.

saadrahim: @pfultz2 The reason --dirty is being removed is CI cannot be changed fast enough for rocm 2.7 to prevent the dirty flag being raised. Jenkins distribution of code from master to slaves is the reason.

aaksud: > You should leave a comment that we will addthe dirty flag again for 2.8.

Ack. Addressed.

aaksud: > I think two PRs is best as one fix and another changes the behavior. The dirty removal will probably be reverted.

Ack.

pfultz2:
Can you split this into two PRs?

share/rocm/cmake/ROCMSetupVersion.cmake
pfultz2: > Maybe a branch for rocm 2.7 is best for this fix. RC candidate branch is the best approach going forward.

I think it would be better to put the removal of "dirty" flag in a separate branch for 2.7.

PR #13, opened 2019-08-01
SOVERSION added to rocm_setup_version

saadrahim:
This allows rocm_setup_version to conveniently manage the SOVERSION for libraries.


pfultz2:
This should have not been merged in. This doesn't set the SOVERSION and even if it did the SOVERSION is not a project-wide setting. I am going to revert this.

PR #15, opened 2019-08-02
rocm_set_soversion added 

saadrahim:
Usable by math libraries to manage the ABI version MAJOR parsing.

### DIFF
--- a/share/rocm/cmake/ROCMSetupVersion.cmake
+++ b/share/rocm/cmake/ROCMSetupVersion.cmake
@@ -158,6 +158,8 @@ endfunction()
 function(rocm_set_soversion)
     set(oneValueArgs TARGET SOVERSION)
 
-    rocm_version_regex_parse("^([0-9]+).*" SOVERSION_MAJOR "${SOVERSION}")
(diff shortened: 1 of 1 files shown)


share/rocm/cmake/ROCMSetupVersion.cmake
```diff
@@ -154,3 +154,12 @@ function(rocm_setup_version)
     endif()
 
 endfunction()
+
+function(rocm_set_soversion)
+    set(oneValueArgs LIBRARY_TARGET SOVERSION)
+
+    cmake_parse_arguments(PARSE "${options}" "${oneValueArgs}" "${multiValueArgs}" ${ARGN})
```
pfultz2: There is actually no need to use `cmake_parse_arguments`. I think doing `function(rocm_set_soversion LIBRARY_TARGET SOVERSION)` would work and should be clear enough when called(ie `rocm_set_soversion(rocfft 0.1)`

saadrahim: Without the parse, both variables were not usable in the function.

saadrahim: I had to add the parse after trying without

saadrahim: I see what you are asking, fixing it.

PR #17, opened 2019-08-06
Variable name changes for rocm_get_build_info

amd-aakash:
Changed the variable names to make them more unique and usable in different contexts.


pfultz2:
If we are using custom env variables then we should have just one variable instead of two.

saadrahim:
Abandoned PR

PR #18, opened 2019-08-06
Swapping the versions for library versioning

saadrahim:


pfultz2:
So #19 has this change plus the tests. Closing this, since the other PR supersedes it.

PR #20, opened 2019-08-08
Set build env identifier from one env variable instead of two.

radhaksri:
Use one environment variable to identify build environment instead of two.


share/rocm/cmake/ROCMSetupVersion.cmake
```diff
@@ -61,11 +61,8 @@ endfunction()
 
 function(rocm_get_build_info OUTPUT DELIM)
     set(_info)
-    if(DEFINED ENV{JOB_NAME})
-        set(_info ${_info}${DELIM}$ENV{JOB_NAME})
-    endif()
-    if(DEFINED ENV{BUILD_ID})
-        set(_info ${_info}${DELIM}$ENV{BUILD_ID})
+    if(DEFINED ENV{ROCM_BUILD_ENV_IDENTIFIER})
```
pfultz2: How about `ROCM_BUILD_ID`?

radhaksri: Done.

saadrahim:
Hi,
Can we get this merged? I really need the builds to stop failing on CI. 
Saad

PR #22, opened 2019-09-13
ROCMCreatePackage support for OS name and architecture

saadrahim:
- CPACK_SYSTEM_NAME is set within rocm_create_package. This is to replace our rpm/debs ending in -Linux.rpm or -Linux.deb. Values are based on /etc/os-release.

- Tested on SLES151, Ubuntu16, Centos7 

Following values are generated for CPACK_SYSTEM_NAME
1. el7.x86_64
2. ubuntu-16.04.amd64
3. sles-15.1.amd64

- rocm_set_os_id is also usable for in cmake to detect the OS. This can replace code that looks for /etc/redhat-release


pfultz2:
Shouldn't these functions be in the `ROCMCreatePackage` module?

Also, how are these functions supposed to be used? I would assume that `rocm_create_package` would call them automatically.

saadrahim:
that makes sense, I will move it to ROCMCreatePackage. 

rocm_set_os_id can be used to cmake statements that are OS specific. See an example in rocBLAS.

https://github.com/ROCmSoftwarePlatform/rocBLAS/blob/ba31b8672094e0807830049343b02cdaa9da405b/clients/CMakeLists.txt#L27-L31


> 
> 
> Shouldn't these functions be in the `ROCMCreatePackage` module?
> 
> Also, how are these functions supposed to be used? I would assume that `rocm_create_package` would call them automatically.

pfultz2:
> rocm_set_os_id can be used to cmake statements that are OS specific. See an example in rocBLAS.

Does [`find_package(OpenMP)`](https://cmake.org/cmake/help/v3.5/module/FindOpenMP.html) not work? It seems very fragile to do this.

saadrahim:
> 
> 
> > rocm_set_os_id can be used to cmake statements that are OS specific. See an example in rocBLAS.
> 
> Does [`find_package(OpenMP)`](https://cmake.org/cmake/help/v3.5/module/FindOpenMP.html) not work? It seems very fragile to do this.

Sadly, find package(OpenMP) does not work. I believe there are bugs with certain cmake versions. I will be refactoring that code in the future as it is really bad. While that is one example, there may be other cases where OS detection is useful.

saadrahim:
I have made the changes as requested. The packaging OS and architecture names are set in RocmCreatePackage. This can be rolled out instantly to all our libraries wit this PR.

I have also broken the CI.  Before I work on fixing the CI and adding some error handling if /etc/os-release does not exist, does everyone agree with naming convention and changes implemented?

share/rocm/cmake/ROCMCreatePackage.cmake
```diff
@@ -98,3 +108,38 @@ macro(rocm_create_package)
     endif()
     include(CPack)
 endmacro()
+
+
+function (rocm_set_system_name)
```
pfultz2: Ok can we have this as `rocm_get_system_name` where the user passes in their own variable? Then the user can decide if they want to set `CMAKE_SYSTEM_NAME` or `CPACK_SYSTEM_NAME` by them setting it manually.

share/rocm/cmake/ROCMCreatePackage.cmake
```diff
@@ -98,3 +108,38 @@ macro(rocm_create_package)
...
+function (rocm_set_os_id OS_ID)
+    set(_os_id "unknown")
+    if (EXISTS "/etc/os-release")
+        rocm_read_os_release("ID" _os_id)
+        endif()
+    set(${OS_ID} ${_os_id} PARENT_SCOPE)
+    set(${OS_ID}_${_os_id} TRUE PARENT_SCOPE)
+endfunction()
+
+function (rocm_read_os_release KEYVALUE OUTPUT)
```
pfultz2: Its common to have the `OUTPUT` variable first in cmake. Maybe this should have the `OUTPUT` variable first?

share/rocm/cmake/ROCMCreatePackage.cmake
```diff
@@ -98,3 +108,38 @@ macro(rocm_create_package)
...
+        STRING(CONCAT _system_name ${_os_id} "-"$ {_version_id} ".amd64")
+    endif()
+    rocm_set_parent(CMAKE_SYSTEM_NAME ${_system_name})
+endfunction()
+
+function (rocm_set_os_id OS_ID)
+    set(_os_id "unknown")
+    if (EXISTS "/etc/os-release")
+        rocm_read_os_release("ID" _os_id)
+        endif()
```
pfultz2: The `end_if` should align with the `if`.

pfultz2:
> I have also broken the CI.

BTW, you can run the tests locally by running `make check`.

share/rocm/cmake/ROCMCreatePackage.cmake
saadrahim: I am removing this as I moved the functionality to rocm_create_package. Do you think it is worth keeping as a separate call?

pfultz2: If you dont think its useful outside of `rocm_create_package` then just remove it.

share/rocm/cmake/ROCMCreatePackage.cmake
```diff
@@ -28,6 +28,16 @@ macro(rocm_create_package)
         set( CPACK_SET_DESTDIR ON CACHE BOOL "Boolean toggle to make CPack use DESTDIR mechanism when packaging" )
     endif()
 
+    rocm_set_os_id(_os_id)
+    rocm_read_os_release("VERSION_ID" _version_id)
+    
+    #only set CPACK_SYSTEM_NAME for AMD supported OSes
+    if (_os_id_centos OR _os_is_rhel)
+        STRING(CONCAT CPACK_SYSTEM_NAME "el" ${_version_id} ".x86_64")
```
pfultz2: `CPACK_SYSTEM_NAME` should be set as a `CACHE` variable, so we can override it if necessary.

share/rocm/cmake/ROCMCreatePackage.cmake
```diff
@@ -28,6 +28,16 @@ macro(rocm_create_package)
...
+    elseif(_os_id_ubuntu OR _os_is_sles)
+        STRING(CONCAT CPACK_SYSTEM_NAME ${_os_id} "-"${_version_id} ".amd64")
```
pfultz2: These should use an underscore instead of `.`. So it should be `_amd64` and `_x86_64`. This is usually common for package names(at least in ubuntu):

```
$ ls /var/cache/apt/archives/  | head
clang-8_1%3a8-3~ubuntu18.04.1_amd64.deb
libclang1-8_1%3a8-3~ubuntu18.04.1_amd64.deb
libclang-common-8-dev_1%3a8-3~ubuntu18.04.1_amd64.deb
libomp5-8_1%3a8-3~ubuntu18.04.1_amd64.deb
libomp-8-dev_1%3a8-3~ubuntu18.04.1_amd64.deb
llvm-8_1%3a8-3~ubuntu18.04.1_amd64.deb
llvm-8-dev_1%3a8-3~ubuntu18.04.1_amd64.deb
llvm-8-runtime_1%3a8-3~ubuntu18.04.1_amd64.deb
```

saadrahim: For Centos, dot seems more common. https://rpmfind.net/linux/rpm2html/search.php?query=gcc&submit=Search+...

I am going to search a little more. 
I went with amd64 for sles and ubuntu as they use that more often. redhat seems to prefere x86_64.

saadrahim: Might have to do something distinct for each OS.

saadrahim: Maybe it is a deb versus RPM convention for period or underscore.

pfultz2: Yea, maybe RPMs use a dot and debians use an underscore.

saadrahim:
Hi Paul,

This does not work. I broke the code when adding the CACHE for CPACK_SYSTEM_NAME. I did’t get a chance to make the fix yesterday. Do you want to revert or give a 1-2 hours to test the fix?

Saad

PR #25, opened 2019-09-25
Changing behaviour on unsupported OS

saadrahim:
I think the default behaviour needs changing. Please let me know what you think.


share/rocm/cmake/ROCMCreatePackage.cmake
```diff
@@ -40,7 +40,8 @@ macro(rocm_create_package)
     elseif(_os_id_sles)
```
saadrahim: Maybe I should remove this as SLES is part of the default case.

share/rocm/cmake/ROCMCreatePackage.cmake
```diff
@@ -37,10 +37,9 @@ macro(rocm_create_package)
     #Debs use underscrore between OS and architecture
     elseif(_os_id_ubuntu)
         STRING(CONCAT _SYSTEM_NAME ${_os_id} "-" ${_version_id} "_amd64")
-    elseif(_os_id_sles)
-        STRING(CONCAT _SYSTEM_NAME ${_os_id} "-" ${_version_id} ".amd64")
     else()
-        set(_SYSTEM_NAME ${CPACK_SYSTEM_NAME})
+    #For SLES and unsupported OSes
+        STRING(CONCAT _SYSTEM_NAME ${_os_id} "-" ${_version_id} ".amd64")
     endif()
```
saadrahim: All modern Linux distributions have /etc/os-release. What should our behaviour be if it does not exist?

pfultz2: We should just use `CMAKE_SYSTEM_NAME` when ` /etc/os-release` is not found.

saadrahim: Please do not merge yet. I fixed it to go with default behavior. I am going to test this out and see.

pfultz2:
@saadrahim Is this ready to merge?

saadrahim:
Please merge, @amdkila tested it.

saadrahim:
HI @pfultz2,
Can you please merge this?

Saad

PR #26, opened 2019-11-07
Updating the rocm_set_soversion function

pruthvistony:
- rocm_set_soverson takes input for version and adds the so patch
  based on the env, if not defined nothing is appended.
- Env variable ROCM_LIBPATCH_VERSION can overwrite the soversion
  patch.
- Added/updated the tests to verify the soversion patch.


share/rocm/cmake/ROCMSetupVersion.cmake
```diff
@@ -154,7 +154,18 @@ endfunction()
 
 function(rocm_set_soversion LIBRARY_TARGET SOVERSION)
     if(NOT WIN32)
-        rocm_version_regex_parse("^([0-9]+).*" SOVERSION_MAJOR "${SOVERSION}")
-        set_target_properties(${LIBRARY_TARGET} PROPERTIES VERSION ${SOVERSION} SOVERSION ${SOVERSION_MAJOR})
+        rocm_version_regex_parse("^([0-9]+).*" LIB_VERSION_MAJOR "${SOVERSION}")
+        rocm_version_regex_parse("^[0-9]+\\.([0-9]+).*" LIB_VERSION_MINOR "${SOVERSION}")
```
pfultz2: This shouldn't chopp off the other versions added in the build.

share/rocm/cmake/ROCMSetupVersion.cmake
```diff
@@ -154,7 +154,18 @@ endfunction()
...
+
+        if (DEFINED ENV{ROCM_LIBPATCH_VERSION})
+            set (LIB_VERSION_PATCH $ENV{ROCM_LIBPATCH_VERSION})
+        else ()
+            set (LIB_VERSION_PATCH 0)
```
pfultz2: If `ROCM_LIBPATCH_VERSION` is not defined then an additional `.` should not be added at all.

share/rocm/cmake/ROCMSetupVersion.cmake
```diff
@@ -154,7 +154,18 @@ endfunction()
 
 function(rocm_set_soversion LIBRARY_TARGET SOVERSION)
     if(NOT WIN32)
-        rocm_version_regex_parse("^([0-9]+).*" SOVERSION_MAJOR "${SOVERSION}")
-        set_target_properties(${LIBRARY_TARGET} PROPERTIES VERSION ${SOVERSION} SOVERSION ${SOVERSION_MAJOR})
+        rocm_version_regex_parse("^([0-9]+).*" LIB_VERSION_MAJOR "${SOVERSION}")
+        rocm_version_regex_parse("^[0-9]+\\.([0-9]+).*" LIB_VERSION_MINOR "${SOVERSION}")
+
+        if (DEFINED ENV{ROCM_LIBPATCH_VERSION})
```
pfultz2: I think a name like `ROCM_SOPATCH_VERSION` would make it clearer this is setting the SO version.

share/rocm/cmake/ROCMSetupVersion.cmake
```diff
@@ -154,7 +154,18 @@ endfunction()
...
+            set (LIB_VERSION_PATCH $ENV{ROCM_LIBPATCH_VERSION})
```
pfultz2: This should call the `rocm_get_build_info` function to append the build information.

test/pass/simple-soversion.cmake
```diff
@@ -12,13 +12,26 @@ install_dir(${TEST_DIR}/libsimple CMAKE_ARGS -DBUILD_SHARED_LIBS=On -DCMAKE_WIND
 if(NOT WIN32 AND NOT APPLE)
     test_expect_file(${PREFIX}/lib/libsimple.so)
     test_expect_file(${PREFIX}/lib/libsimple.so.1)
-    test_expect_file(${PREFIX}/lib/libsimple.so.1.1.2)
+    if (DEFINED ENV{ROCM_LIBPATCH_VERSION})
+        test_expect_file(${PREFIX}/lib/libsimple.so.1.1.$ENV{ROCM_LIBPATCH_VERSION})
```
pfultz2: The test shouldn't check for the ENV variable. It should set the `ENV` variable and then verify the so version was set correctly.

share/rocm/cmake/ROCMSetupVersion.cmake
pruthvistony: This name is being set in the jenkins build environment and being agreed upon by other teams.
One thing I can do is to set ROCM_SOPATCH_VERSION as cache variable and pass the value as a parameter while building the package. Let me know if this fine.

share/rocm/cmake/ROCMSetupVersion.cmake
pruthvistony: Ok. I will update the changes and call the rocm_get_build_info to the ROCM_LIBPATCH_VERSION from env.

share/rocm/cmake/ROCMSetupVersion.cmake
pruthvistony: The idea with the patch was to add ".<patch-ver>" to the soname name always. Eg: if ROCM_LIBPATCH_VERSION is defined then ".20901" or ".0" since default is 0. I believe there is no case where only "." is added. Please clarify on this.

test/pass/simple-soversion.cmake
pruthvistony: I can set the env variable in the simple-soversion.cmake and verify the test. However the approach is dependent on the previous comment on how to pass the ROCM_LIBPATCH_VERSION as a env or cmake command line parameter.

share/rocm/cmake/ROCMSetupVersion.cmake
pruthvistony: I have updated the changes not to chop off the other versions.

pruthvistony: Please review, I have updated the changes.

share/rocm/cmake/ROCMSetupVersion.cmake
pruthvistony: Please review the latest commit.

share/rocm/cmake/ROCMSetupVersion.cmake
pruthvistony: I have updated the way to access the ROCM_LIBPATCH_VERSION though the rocm_get_so_patch function. Please check the latest commit.

test/pass/simple-soversion.cmake
pruthvistony: I have updated to tests to remove the ENV. Added separate tests to verify the operation by setting the ENV variables. Please check the latest commit.

pruthvistony:
Please check the latest commit and let me know your comment.

share/rocm/cmake/ROCMSetupVersion.cmake
```diff
@@ -152,9 +152,33 @@ function(rocm_setup_version)
...
+
+        set_target_properties(${LIBRARY_TARGET} PROPERTIES SOVERSION ${LIB_VERSION_MAJOR} )
+        set_target_properties(${LIBRARY_TARGET} PROPERTIES VERSION ${LIB_VERSION_STRING} )
+    endif()
+endfunction()
+
+function(rocm_set_runpath)
+    if(NOT WIN32 AND NOT APPLE)
+        if(${ARGC} GREATER 0)
+            set ( CMAKE_SHARED_LINKER_FLAGS " -Wl,--enable-new-dtags -Wl,--rpath,${ARGV0}" PARENT_SCOPE )
```
pfultz2: These should be set by the toolchain not by rocm-cmake. We should never be overriding these variables.

test/pass/simple-soversion.cmake
```diff
@@ -12,13 +12,13 @@ install_dir(${TEST_DIR}/libsimple CMAKE_ARGS -DBUILD_SHARED_LIBS=On -DCMAKE_WIND
 if(NOT WIN32 AND NOT APPLE)
     test_expect_file(${PREFIX}/lib/libsimple.so)
     test_expect_file(${PREFIX}/lib/libsimple.so.1)
-    test_expect_file(${PREFIX}/lib/libsimple.so.1.1.2)
+    test_expect_file(${PREFIX}/lib/libsimple.so.1.1.0)
```
pfultz2: These tests shouldn't change if the `ROCM_LIBPATCH_VERSION` env variable is not set.

test/pass/simple-soversion-env.cmake
```diff
@@ -0,0 +1,28 @@
...
+    test_expect_file(${PREFIX}/lib/libsimple.so.1)
+    test_expect_file(${PREFIX}/lib/libsimple.so.1.1.$ENV{ROCM_LIBPATCH_VERSION})
+
+    test_expect_realpath(${PREFIX}/lib/libsimple.so ${PREFIX}/lib/libsimple.so.1.1.$ENV{ROCM_LIBPATCH_VERSION})
+    test_expect_realpath(${PREFIX}/lib/libsimple.so.1 ${PREFIX}/lib/libsimple.so.1.1.$ENV{ROCM_LIBPATCH_VERSION})
+
+    test_exec(COMMAND ldd ${PREFIX}/bin/simple-main OUTPUT_VARIABLE LIBS)
+    test_expect_matches("${LIBS}" "libsimple.so.1 =>")
+    test_expect_not_matches("${LIBS}" "libsimple.so =>")
+    test_expect_not_matches("${LIBS}" "libsimple.so.1.1.$ENV{ROCM_LIBPATCH_VERSION} =>")
```
pfultz2: This should be `1.1.2.$ENV{ROCM_LIBPATCH_VERSION}`. The version should not be stripped off of the the so file.

share/rocm/cmake/ROCMSetupVersion.cmake
pruthvistony: Sorry I do get what you mean by toolchain. Is it that we should never have a api in rocm cmake to set the runpath? Can you please clarify.

test/pass/simple-soversion.cmake
pruthvistony: As per the code the behavior is if ROCM_LIBPATCH_VERSION is not set, it defaults the patch version to 0 and appends ".0" to the so file name. So suggestions is to not added any patch version is it?

test/pass/simple-soversion-env.cmake
pruthvistony: I will update the patch accordingly.

pruthvistony:
Please check my response to the comments, I need few clarifications. 
Thank you.

test/pass/simple-soversion.cmake
pfultz2: >  So suggestions is to not added any patch version is it?

Nope, the so version should be exactly what is specified by `rocm_set_soversion` function. If `ROCM_LIBPATCH_VERSION` is defined then that should append the version defined by `rocm_set_soversion` with `.${ROCM_LIBPATCH_VERSION}`.

share/rocm/cmake/ROCMSetupVersion.cmake
pfultz2: > Is it that we should never have a api in rocm cmake to set the runpath?

More than that, it should never appear in the CMakeLists.txt ever. Instead this should be passed in when running cmake either by passing `-DCMAKE_SHARED_LINKER_FLAGS=...` or putting it in a cmake toolchain file and passing the file in with `-DCMAKE_TOOLCHAIN_FILE=...`.

pfultz2: Also, if you are using `rbuild` the `-D` variables can be passed to it as well or you can pass the cmake toolchain file with `--toolchain` flag.

test/pass/simple-soversion.cmake
pruthvistony: I have updated the latest commit as per the above comment.

share/rocm/cmake/ROCMSetupVersion.cmake
pruthvistony: I understand. I will remove this function (rocm_set_runpatch) and 2 test cases which check this api.

pruthvistony:
Please review. 
I have updated the commit with all the review comments taken care.

PR #27, opened 2019-11-11
Fix typo for Red Hat detection

eidenyoshida:


eidenyoshida:
For ticket SWDEV-210022

PR #28, opened 2019-12-02
Fix OpenSUSE string to match SLES packaging naming convention

amdkila:
OpenSUSE distros such as Leap have OS-ID of opensuse-leap instead of sles, thus we should map this ID to sles or else there will be packaging issues for OpenSUSE users.


share/rocm/cmake/ROCMCreatePackage.cmake
```diff
@@ -121,6 +121,9 @@ function (rocm_set_os_id OS_ID)
     if (EXISTS "/etc/os-release")
         rocm_read_os_release(_os_id "ID")
     endif()
+    if (_os_id STREQUAL "opensuse-leap")
```
pfultz2: Would it be better to write `_os_id MATCHES "opensuse"`?

amdkila:
Should merge PR #25 as it provides a better solution.

PR #29, opened 2019-12-16
Controlling the ld.conf update in create_package

pruthvistony:
Controlling the ld.conf update in rocm_create_package macro
    
Default ROCM_DISABLE_LDCONFIG is set to OFF, can be overridden
as a cmake toolchain parameter.


pruthvistony:
@pfultz2 
Please review this PR. I have verified the behavior manually checking the package.
Not sure how this can be added into a test case.

pfultz2:
There should be a cmake flag `ROCM_DISABLE_LDCONFIG` that is checked instead:

```cpp
set(ROCM_DISABLE_LDCONFIG Off BOOL CACHE "")

macro(rocm_create_package)
...

    if(PARSE_LDCONFIG AND NOT ${ROCM_DISABLE_LDCONFIG})
        ...
    endif()
endmarco()
```

pruthvistony:
Updated the code as per the review comment. Please review.

PR #31, opened 2020-04-02
Print WARNING instead of ERROR for modified toolchain variables

leekillough:
Modified toolchain variables should print a `WARNING`, not an `ERROR`, since they are non-fatal, and since several projects have legacy CMake code which modifies toolchain variables, and which are unlikely to be fixed soon.

Printing `ERROR` makes it hard to find real errors in builds, such as by searching for "error" strings in logs.

The warnings can continue to be loud and prominent, in order to encourage the fixing of legacy code which violates toolchain conventions.

@amcamd @saadrahim @amdkila @TorreZuk @eidenyoshida @bragadeesh @sdquiring @zaliu

### DIFF
--- a/share/rocm/cmake/ROCMChecks.cmake
+++ b/share/rocm/cmake/ROCMChecks.cmake
@@ -16,7 +16,7 @@ function(rocm_check_toolchain_var var access value)
     if(access STREQUAL "MODIFIED_ACCESS")
         message("
 *******************************************************************************
-*----------------------------------- ERROR -----------------------------------*
+*---------------------------------- WARNING ----------------------------------*
 * The variable '${var}' should only be set by the cmake toolchain,
 * either by calling 'cmake -D${var}=\"${value}\"' or
 * set in a toolchain file and added with 


pfultz2:
The plan is to change this to a fatal error, which is why it says `ERROR` so people will treat them as errors and fix the issue. It isn't an error yet, but this is to give developers time to fix any issue or to work out any issues with the checks. I dont think we should change the message.

Currently, there is an issue where it will print these messages if the `project` command is called after `find_package(ROCM)`, which I hope to fix.

TorreZuk:
I also think WARNING is better, we have been confusing users for many months when they see ERROR as cmake functions as expected, it is just bad practice.    Should I make a PR to fix the erroneous printing of this message for each project( ) in subdirectories ?     I'll just match the right side path to be a cmake output path?   @pfultz2 is a PR to master the correct branch?

TorreZuk (approved):

pfultz2:
I would like to make this an actual error so this will be caught in the CI, we just need to make sure there is no false positives. It doesn't make sense to print WARNING if its an error.

> Should I make a PR to fix the erroneous printing of this message for each project( ) in subdirectories ?

Yes, it checks for some cmake directories [here](https://github.com/RadeonOpenCompute/rocm-cmake/blob/master/share/rocm/cmake/ROCMChecks.cmake#L24) but it seems we have a few more cases where it isnt caught yet.

TorreZuk:
Yes I know your intent but it isn't an ERROR right now and as you say there are false positives ( I get them from L27 ).
I still think it needs to be changed to WARNING until there are **zero** false positives and all teams should agree to make it an error when we schedule time to fix the failures.   I'll make a PR to try and fix some of the failures.

TorreZuk:
@pfultz2 as is covered by merged https://github.com/RadeonOpenCompute/rocm-cmake/pull/62 so expect @leekillough  is fine if we close this.

PR #33, opened 2020-04-03
Make the names for packages lowercase

pfultz2:


daniellowell:
We should probably be bumping the version on this PR.

PR #34, opened 2020-04-24
Add clang tidy target to check all sources related to the original target

tafryn:
This provides a missing middle-ground between running clang-tidy on the entire project, and running clang-tidy on individual source files.


pfultz2:
Thanks!

PR #35, opened 2020-05-19
Adding libpatch version to VERSION for uniform package naming structure

ashutom:
As per new naming standard :
Version=Major.minor.patch.<libpatch>

This change aims to execute the same.


saadrahim:
Abandoned PR

PR #42, opened 2021-01-08
Add option to disable ROCM Toolchain check completely

davidrohr:
By default ROCm CMake will always print a message when a toolchain variable was modified, even if the warning and the error were switched off explicitly. This adds another switch to disable the message completely.

### DIFF
--- a/share/rocm/cmake/ROCMChecks.cmake
+++ b/share/rocm/cmake/ROCMChecks.cmake
@@ -8,6 +8,9 @@ set(ROCM_WARN_TOOLCHAIN_VAR
 set(ROCM_ERROR_TOOLCHAIN_VAR
     OFF
     CACHE BOOL "")
+set(ROCM_CHECK_TOOLCHAIN_VAR
+    ON
+    CACHE BOOL "")
 
 function(rocm_check_toolchain_var var access value list_file)
     set(message_type STATUS)
@@ -26,7 +29,7 @@ function(rocm_check_toolchain_var var access value list_file)
         elseif("${base}" MATCHES ".*/CMakeFiles/${CMAKE_VERSION}$")
             set(cmake_module On)
         endif()
-        if(NOT cmake_module)
+        if(NOT cmake_module AND ROCM_CHECK_TOOLCHAIN_VAR)
             message(
                 "
 *******************************************************************************
--- a/share/rocm/cmake/ROCMChecks.cmake
+++ b/share/rocm/cmake/ROCMChecks.cmake
@@ -13,25 +13,26 @@ set(ROCM_CHECK_TOOLCHAIN_VAR
     CACHE BOOL "")
 
 function(rocm_check_toolchain_var var access value list_file)
-    set(message_type STATUS)
-    if(ROCM_WARN_TOOLCHAIN_VAR)
-        set(message_type WARNING)
-    endif()
-    if(ROCM_ERROR_TOOLCHAIN_VAR)
-        set(message_type SEND_ERROR)
-    endif()
-    if(access STREQUAL "MODIFIED_ACCESS")
-        set(cmake_module Off)
-        get_filename_component(base "${list_file}" DIRECTORY)
-        # Skip warning in cmake's built-in modules
-        if("${base}" STREQUAL "${CMAKE_ROOT}/Modules")
-            set(cmake_module On)
-        elseif("${base}" MATCHES ".*/CMakeFiles/${CMAKE_VERSION}$")
-            set(cmake_module On)
+    if (ROCM_CHECK_TOOLCHAIN_VAR)
+        set(message_type STATUS)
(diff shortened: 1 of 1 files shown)


pfultz2:
We should fix the issues with modifying toolchain variables instead of disabling this completely.

share/rocm/cmake/ROCMChecks.cmake
```diff
@@ -26,7 +29,7 @@ function(rocm_check_toolchain_var var access value list_file)
         elseif("${base}" MATCHES ".*/CMakeFiles/${CMAKE_VERSION}$")
             set(cmake_module On)
         endif()
-        if(NOT cmake_module)
+        if(NOT cmake_module AND ROCM_CHECK_TOOLCHAIN_VAR)
             message(
```
TorreZuk, line 33: The short circuit should go earlier on the access modified test as no need to do the work detecting if cmake itself making the modification if we aren't messaging.   So:
if (access STREQUAL "MODIFIED_ACCESS" AND ROCM_CHECK_TOOLCHAIN_VAR)

TorreZuk (approved):
I agree in many cases you have no control over the code as it is frozen and you know it works correctly and don't want to see all these messages as they often break systems that grep for ERROR.

davidrohr:
> break systems that grep for ERROR

That was exactly my use case. I have moved the check to an earlier location as requested.

saadrahim (approved):

pfultz2:
> you have no control over the code as it is frozen and you know it works correctly

We have control over the projects that are using rocm-cmake. These errors should be fixed in those projects. If its a false positive, then it should be fixed in rocm-cmake.

> they often break systems that grep for ERROR.

I plan to make this a hard error in cmake anyways so it wont be ignored by projects.

davidrohr:
> We have control over the projects that are using rocm-cmake. These errors should be fixed in those projects. If its a false positive, then it should be fixed in rocm-cmake.

I am afraid I cannot really agree with this. When my project uses 2 external packages, hip and another one, and I do in cmake
```
find_package(hip)
find_package(xxx)
```
and `xxx` modifies CMAKE_CXX_FLAGS, this will trigger an error, while I might have no control over `xxx`. I think if modification to toolchain variables are to be prohibited, this should be on the CMake level, but I have not seen where CMake forbids it.

pfultz2:
> When my project uses 2 external packages, hip and another one, and I do in cmake

`find_package(hip)` should not include rocm-cmake. If thats the case, then it should be removed from `find_package(hip)`. rocm-cmake modules are meant to be a build-only dependency, and downstream projects should never be required to use it.

>  I think if modification to toolchain variables are to be prohibited, this should be on the CMake level, but I have not seen where CMake forbids it.

CMake doesnt prohibit it, but we want to prohibit this in the build scripts of rocm packages as modifying these variables is problematic because it ignores what the user is requesting.

davidrohr:
OK, thx, that explains something. So I have ROCm 4.0, and I use
```
    find_package(hip)
    find_package(hipcub)
    find_package(rocprim)
    find_package(rocthrust)
```
which pulls in the ROCmChecks.cmake for me. If that is not intended, I agree then it should be fixed there.

davidrohr:
Could you comment what is the status? The travis CI fails, but I do not understand why.
Do you plan to merge this, or is it planed to solve this in a different way in which case I could close it.

pfultz2:
I think the plan is to figure out which package is including rocm-cmake. Let me see if I can get a system with rocm 4.0 so I can try and grep for `ROCM`.

I am not sure what the issue is with travis, but its not a required status so can be ignored. The other stages need to run, I think you need to merge in master for it to run.

pfultz2:
Grepping over our rocm 4.0 packages doesn't show any include rocm-cmake. Perhaps your project is pulling this in from somewhere else. Is the code for this project available on github?

pfultz2:
I also created a simple cmake with rocm 4.0:

```cmake
cmake_minimum_required(VERSION 3.10)
project(sample)

list(APPEND CMAKE_PREFIX_PATH /opt/rocm/hip /opt/rocm)

find_package(hip)
find_package(hipcub)
find_package(rocprim)
find_package(rocthrust)

set(CMAKE_CXX_FLAGS "-std=c++11")
```

I dont see an error from the ROCMChecks, but if I explicitly add `find_package(ROCM)` then these errors will show.

davidrohr:
@pfultz2 : Thx a lot, it is indeed the `find_package(ROCM)`. We have a HIP and an OpenCL backend, the HIP was actually not doing it, but the OpenCL backend did a `find_package(ROCM)` in order to probe for the `clang-ocl` script. I have changed that to a hardcoded patch `/opt/rocm/bin` and removed the `find_package(ROCM)`, and now the error is gone. Thx, closing this. But let me just remark that it seems quite misleading to me that a `find_package(ROCM)` yields such errors, perhaps that could be fixed in any case.

PR #43, opened 2021-04-09
Add rocm_check_target_ids

cgmb:
I've added a function that will check if the given list of target-ids are supported by the current CXX compiler, and return the subset that are.

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,3 +107,15 @@ rocm_setup_version
     )
 
 Setup the version for the project. This will try to use git tag to set the version if possible unless `NO_GIT_TAG_VERSION` is passed. The `PARENT` argument can be used to set the commit to start the count of number of commits to the current revision.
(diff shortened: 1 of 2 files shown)


saadrahim:
Can the version number of rocm-cmake be incremented for us to check the availability of this feature?

cgmb:
@saadrahim, [include](https://cmake.org/cmake/help/latest/command/include.html) supports `OPTIONAL` and `RESULT_VARIABLE <variable>`, so I think you can check for the presence of this feature with:
```
include(ROCMCheckTargetIds OPTIONAL RESULT_VARIABLE HAVE_ROCM_CHECK_TARGET_IDS)
if(HAVE_ROCM_CHECK_TARGET_IDS)
  # here you know you have rocm_check_target_ids
endif()
```

PR #51, opened 2021-07-19
Automatically setup devel packages

lawruble13:
This is a proposed solution to set up devel packages for all libraries with minimal effort.

In order to enable the devel packaging, library maintainers would need to call the function `rocm_set_devel_component()` prior to the first call to `rocm_install`, `rocm_install_targets`, `rocm_export_targets`, or `rocm_install_symlink_subdir`. Once `rocm_set_devel_component` is called, the aforementioned functions will install everything that would have been installed to the base package (except library binaries) to the devel package instead. Library binaries are namelinked in the devel package, and are installed to the base package. Any items which are installed using
- A) `install`;
- B) `rocm_install` with the added arguments `COMPONENT Unspecified`; or
- C) `install` or `rocm_install` prior to the call to `rocm_set_devel_component`
will install to the base package.


share/rocm/cmake/ROCMConfigVersion.cmake
```diff
@@ -0,0 +1,51 @@
...
+# to create a version-file which can be installed along a config.cmake file.
+#
+# The created file sets PACKAGE_VERSION_EXACT if the current version string and
+# the requested version string are exactly the same and it sets
+# PACKAGE_VERSION_COMPATIBLE if the current version is >= requested version,
+# but only if the requested major version is the same as the current one.
+# The variable CVF_VERSION must be set before calling configure_file().
+
+
+set(PACKAGE_VERSION "0.6.0")
```
pfultz2, line 12: This seems like a duplicate to what `write_basic_package_version_file` generates. Why do we have this defined twice? Also the version should be defined in CMakeLists.txt not in this file as well(we should have one place to define the version).

CHANGELOG.md
```diff
@@ -1,14 +1,35 @@
 # Change Log for rocm-cmake
 
+## [0.6.0]
+### Added
+- `rocm_set_devel_component` can be used to set the component used for devel packaging (defaults to devel).
```
pfultz2, line 5: I think it would be better to have an opt-out instead of an opt-in setting. I think it would be better to have a global cached variable such as `ROCM_ENABLE_DEV_COMPONENTS`, and libraries that wont work with this can either set it to off before calling `find_package(ROCM)` or when calling cmake(ie `cmake -DROCM_ENABLE_DEV_COMPONENTS=Off`).

share/rocm/cmake/ROCMConfigVersion.cmake
lawruble13: ROCMConfigVersion.cmake is the file generated by `write_basic_package_version_file`, included here only for convenience in automatic downloads of rocm-cmake (as used by almost all math libraries). By including the version file, the version can be checked and the files included directly from the download. However, it does duplicate the version information and is unrelated to the purpose of this PR and could be removed.

share/rocm/cmake/ROCMInstallTargets.cmake
```diff
@@ -7,10 +7,55 @@ include(GNUInstallDirs)
...
+        set(COMPONENT "devel")
+    else()
+        set(COMPONENT "${ARGV0}")
+    endif()
+    if(DEFINED CACHE{DEVEL_COMPONENT})
+        if(NOT "$CACHE{DEVEL_COMPONENT}" STREQUAL ${COMPONENT})
+            message(WARNING "Development component being reset from $CACHE{DEVEL_COMPONENT} to ${COMPONENT}.")
+        endif()
+    endif()
+    set(DEVEL_COMPONENT "${COMPONENT}" CACHE INTERNAL "The component to use for a devel package.")
```
pfultz2, line 23: Global variables should be prefixed with `ROCM_`.

share/rocm/cmake/ROCMInstallTargets.cmake
```diff
@@ -7,10 +7,55 @@ include(GNUInstallDirs)
...
+            list(APPEND ACTUAL_COMMAND FILES_MATCHING "${PARSE_FILES_MATCHING}")
+        endif()
+        install(${ACTUAL_COMMAND})
+    endif()
+endfunction()
 
 function(rocm_install_targets)
     set(options)
-    set(oneValueArgs PREFIX EXPORT)
+    set(oneValueArgs PREFIX EXPORT COMPONENT)
```
pfultz2, line 58: The README should be updated with the new `COMPONENT` argument.

CHANGELOG.md
lawruble13: I agree that is a nicer solution, the justification behind having an opt-in system is to maintain backwards compatibility. Opt-out would be a breaking change in behaviour, so anyone building for older versions of ROCm would have issues. That said, I'm not sure how much of a concern backwards compatibility is.

share/rocm/cmake/ROCMConfigVersion.cmake
pfultz2: > included here only for convenience in automatic downloads of rocm-cmake (as used by almost all math libraries)

But the math libraries should do install to get the version file. 

> However, it does duplicate the version information

The version should be defined in one place. We could define the version in ROCMVersion.cmake and include that in the CMakeLists.txt and ROCMVersionConfig.cmake can include it. 

> and is unrelated to the purpose of this PR and could be removed.

It would be better to put this a seperate PR.

CHANGELOG.md
pfultz2: With a cached variable, the user can just pass `-DROCM_ENABLE_DEV_COMPONENTS=Off` to build an older library with a newer version of rocm-cmake.

share/rocm/cmake/ROCMInstallTargets.cmake
```diff
@@ -7,10 +7,55 @@ include(GNUInstallDirs)
...
+    endif()
+    if(DEFINED CACHE{DEVEL_COMPONENT})
+        if(NOT "$CACHE{DEVEL_COMPONENT}" STREQUAL ${COMPONENT})
+            message(WARNING "Development component being reset from $CACHE{DEVEL_COMPONENT} to ${COMPONENT}.")
+        endif()
+    endif()
+    set(DEVEL_COMPONENT "${COMPONENT}" CACHE INTERNAL "The component to use for a devel package.")
+endfunction()
+
+function(rocm_install)
```
pfultz2, line 26: This new function should be added to the README file.

share/rocm/cmake/ROCMInstallTargets.cmake
```diff
@@ -44,10 +101,22 @@ function(rocm_install_targets)
         target_include_directories(${TARGET} INTERFACE $<INSTALL_INTERFACE:$<INSTALL_PREFIX>/include>)
     endforeach()
 
+    if(PARSE_COMPONENT)
+        set(COMPONENT_ARG "COMPONENT;${PARSE_COMPONENT}")
+        set(LIB_COMPONENT_ARG "COMPONENT;${PARSE_COMPONENT}")
+    elseif(ROCM_USE_DEV_COMPONENT)
+        set(COMPONENT_ARG "COMPONENT;devel")
+        set(LIB_COMPONENT_ARG "NAMELINK_COMPONENT;devel")
```
pfultz2, line 109: `NAMELINK_COMPONENT` is not an available option in ubuntu 18.04. You should check the cmake version is at least 3.12 otherwise just fallback on this being in the runtime component.

lawruble13: Alternatively, the older method of splitting the namelink and library files (using `NAMELINK_ONLY` and `NAMELINK_SKIP`, which are supported at least back to CMake 3.0) is still supported in later versions of CMake, so it can be used for all platforms.

lawruble13:
@pfultz2 In order to pass the linter, the `max_statements` value was increased. I'm happy to open an issue and revisit that soon, but this as this needs to be approved as soon as possible in order to proceed with getting all of the math libraries compliant, I would ask that the PR be approved despite this.

CMakeLists.txt
```diff
@@ -9,13 +9,13 @@ set(CMAKE_INSTALL_PREFIX "/opt/rocm" CACHE PATH "")
 project(rocm-cmake)
 
 install(DIRECTORY share DESTINATION .)
-install(FILES ${CMAKE_CURRENT_BINARY_DIR}/ROCMConfigVersion.cmake DESTINATION share/rocm/cmake)
+install(FILES ${CMAKE_CURRENT_BINARY_DIR}/ROCMConfigVersion.cmake DESTINATION "share/rocm/cmake")
```
pfultz2, line 12: Is the quotes really necessary?

share/rocm/cmake/ROCMInstallTargets.cmake
```diff
@@ -56,13 +123,50 @@ function(rocm_install_targets)
...
     endforeach()
 
-    install(
-        TARGETS ${PARSE_TARGETS}
-        EXPORT ${EXPORT_FILE}
-        RUNTIME DESTINATION ${BIN_INSTALL_DIR}
-        LIBRARY DESTINATION ${LIB_INSTALL_DIR}
-        ARCHIVE DESTINATION ${LIB_INSTALL_DIR})
-
+    if(${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.12.0")
```
pfultz2, line 126: Is this necessary? Can the implementation for 3.10 work on newer cmake? Or is there a functional difference?

CMakeLists.txt
lawruble13: No, they're not. When removing the ROCMConfigVersion.cmake file, I had to add this line back and added the quotes out of habit. They serve no additional purpose.

share/rocm/cmake/ROCMInstallTargets.cmake
saadrahim: Devops is on cmake 3.18. Do we need to support older cmake versions?

share/rocm/cmake/ROCMInstallTargets.cmake
```diff
@@ -7,10 +7,67 @@ include(GNUInstallDirs)
...
+        install(${ARGN})
+    else()
+        # We want to define the COMPONENT argument in the correct position, only if the user did not define
+        #  the COMPONENT argument. Therefore, capture the component argument and any arguments which can
+        #  legally follow it, so we can place those after the inserted COMPONENT argument.
+        set(options OPTIONAL EXCLUDE_FROM_ALL)
+        set(oneValueArgs COMPONENT RENAME)
+        # Specifying all valid first arguments as multiValueArgs captures all arguments between the first argument
+        #  and the COMPONENT argument (or any argument which can follow COMPONENT) in order.
+        set(multiValueArgs FILES PROGRAMS DIRECTORY CODE SCRIPT EXPORT FILES_MATCHING)
```
pfultz2, line 31: Shouldn't `PROGRAMS` be in the runtime component not dev component?

share/rocm/cmake/ROCMInstallTargets.cmake
lawruble13: The 3.10 implementation apparently does work for newer CMake, previously I had encountered errors complaining about having a target multiple times in one export file when running the test suite on newer CMake versions. Reordering the installation of NAMELINK_SKIP and NAMELINK_ONLY apparently fixed this issue.

aak-amd: We are on cmake 3.16.8 as of now:

docker run -it --rm compute-artifactory.amd.com:5000/rocm-plus-docker/ubuntu-18.04-bld:latest bash
root@7b027ce1e7bb:/src# cmake --version
cmake version 3.16.8

CMake suite maintained and supported by Kitware (kitware.com/cmake).

pfultz2: First, to pass CI it needs to work with cmake 3.10 or 3.20(the versions in ubuntu 18.04 and 20.04 respectively). Secondly, for components like MIOpen and MIGraphX, we build with the cmake that comes with ubuntu 18.04, so we need to make sure it works with at least 3.10.

The question is if the 3.12 implementation is necessary as we could just use the implementation that works for 3.10? But maybe there is a functional difference.

pfultz2: > Reordering the installation of NAMELINK_SKIP and NAMELINK_ONLY apparently fixed this issue.

So it seems it would be simpler to just use the 3.10 implementation instead of having two different versions here.

CMakeLists.txt
pfultz2: I see.

share/rocm/cmake/ROCMInstallTargets.cmake
lawruble13: I concur, just running a couple of further local tests to double check that the 3.10 implementation is functionally identical.

share/rocm/cmake/ROCMInstallTargets.cmake
lawruble13: This is a good point. This also applies to `RUNTIME` (i.e. executable) items installed using `rocm_install_targets`, which should probably go in the runtime component rather than the dev component. Some libraries may generate executables that are only used for developing though, I will double check on that.

lawruble13: This have both been implemented, although this may also need to be revisited in the future.

pfultz2 (approved):

lawruble13:
Hi @pfultz2, can you merge this? It will let us proceed with testing

lawruble13:
@pfultz2 Sorry to keep bothering you, but we really need this merged in order for the devel packages to be tested on CI.

PR #52, opened 2021-07-23
Fix invalid "Suggests:" field on old rpmbuild

lawruble13:
Unfortunately this was missed in earlier testing, the "Suggests:" field used to smooth transitioning is not available on versions of rpmbuild prior to 4.12.0, which caused builds on Centos 7 to not generate the runtime package due to the invalid field. 

To pass the linter, the number of allowed branches was increased from 20 to 25.


saadrahim (approved):
Minor change to check rpmbuild version should have no impact overall.

PR #53, opened 2021-07-27
Update for backwards compatibility

lawruble13:
Fix issues breaking backwards compatibility:
- Issue: CMP0057 not set to NEW
  - Fix: `ROCMInstallTargets.cmake` specifies `cmake_minimum_required(VERSION 3.6)`, which sets CMP0057 to NEW
- Issue: TGZ generator `unable to read from file <CPACK_PACKAGING_INSTALL_PREFIX><CMAKE_INSTALL_PREFIX>/path/to/file`
  - Fix: When `CPACK_SET_DESTDIR` is set, the DEB and RPM generators ignore `CPACK_PACKAGING_INSTALL_PREFIX` anyway, so it is unset, which fixes the TGZ generator

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Change Log for rocm-cmake
 
+## [0.6.1]
+### Fixed
+- Add `cmake_minimum_required` check to `ROCMInstallTargets.cmake` to fix error with `CMP0057` not set
+- Unset `CPACK_PACKAGING_INSTALL_PREFIX` when `CPACK_USE_DESTDIR` is set to fix error with TGZ generator
+
 ## [0.6.0]
 ### Added
 - The cache variable `ROCM_USE_DEV_COMPONENT` may be set to `OFF` to build with legacy behaviour.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,5 @@
 # Change Log for rocm-cmake
 
-## [0.6.1]
-### Fixed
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Change Log for rocm-cmake
 
+## [0.6.1]
+### Fixed
+- Add `cmake_minimum_required` check to `ROCMInstallTargets.cmake` to fix error with `CMP0057` not set
+- Unset `CPACK_PACKAGING_INSTALL_PREFIX` when `CPACK_USE_DESTDIR` is set to fix error with TGZ generator
```
saadrahim, line 6: Not sure this needs to count as a separate version, there has been no release 0.6.0

lawruble13:
Also fix issue with regular expression not being escaped properly (replace `\.` with `\\.` in regular expression string).

saadrahim (approved):
@pfultz2 Please merge as soon as you are satisfied. There are deadlines approaching that need this fix.

share/rocm/cmake/ROCMInstallTargets.cmake
```diff
@@ -2,6 +2,8 @@
 # Copyright (C) 2017 Advanced Micro Devices, Inc.
 # ######################################################################################################################
 
+cmake_minimum_required( VERSION 3.6 )
```
pfultz2, line 5: I think it would be better to just set the policy needed.

lawruble13: Changed to only set CMP0057 to NEW.

PR #54, opened 2021-07-30
Fix packaging issues

lawruble13:
- Issue: When `CPACK_PACKAGING_INSTALL_PREFIX` is specified on the command line, it is not cleared by `unset` in `ROCMCreatePackage.cmake`. This causes packaging to fail for the `TGZ` generator.
  - Fix: If `CPACK_SET_DESTDIR` is used, then obscure any previous definition of `CPACK_PACKAGING_INSTALL_PREFIX` with a local empty string.
  - New behaviour: If the library CMake scripts and/or the user don't clear `CPACK_SET_DESTDIR`, then `CMAKE_INSTALL_PREFIX` is used for all generators. If `CPACK_SET_DESTDIR` is intentionally cleared by the scripts or user, then assume they've set up `CPACK_PACKAGING_INSTALL_PREFIX` accordingly. 
- Issue: Header only libraries (such as rocPRIM) have no contents, and the same dependencies as many other libraries, so they end up being "totally replaced" by other libraries and thus not installed. This breaks many dependencies.
  - Fix: Add a `HEADER_ONLY` option to `rocm_create_package`, which will set the devel package to provide the appropriate runtime package, as well as pick up any dependencies the package was specified to have. Add a `HEADER_ONLY` parameter to `rocm_set_comp_cpackvar` to prevent the `Unspecified` component from generating a package.
  - New behaviour: If the HEADER_ONLY option is provided, only the -devel package will be generated with the appropriate dependency fields and "Provides" field populated. Otherwise, the behaviour is the same.


lawruble13:
@pfultz2 Can you have a look at this? It's required to fix issues with the meta package rocm-libs.

saadrahim (commented):
Please document the new rocm_create_package option for HEADER_ONLY.

saadrahim (approved):

pfultz2:
> : When CPACK_PACKAGING_INSTALL_PREFIX is specified on the command line, it is not cleared by unset

Yea of course because its a cached variable. The `unset` should probably have never been added.

> This causes packaging to fail for the TGZ generator.

What exactly fails here?

> then obscure any previous definition of CPACK_PACKAGING_INSTALL_PREFIX with a local empty string.

Why not set `CPACK_PACKAGING_INSTALL_PREFIX` to `CMAKE_INSTALL_PREFIX` when DESTDIR is set?

> Issue: Header only libraries (such as rocPRIM) have no contents, and the same dependencies as many other libraries, so they end up being "totally replaced" by other libraries and thus not installed. This breaks many dependencies. 

I am not sure I follow this. What do you mean "totally replaced"?

> Fix: Add a HEADER_ONLY option to rocm_create_package, which will set the devel package to provide the appropriate runtime package

But if its a header only then it should only be a dev package. I dont understand why it would provide a runtime package.

lawruble13:
> Yea of course because its a cached variable. The `unset` should probably have never been added.
> ...
> What exactly fails here? As this seems to work before we added the component packaging.

When component packaging is enabled, the `TGZ` generator uses both `CPACK_PACKAGING_INSTALL_PREFIX` and `CMAKE_INSTALL_PREFIX`, so an error is encountered with something like `cannot lstat '/opt/rocm/opt/rocm'` when `CPACK_PACKAGING_INSTALL_PREFIX` is set to `CMAKE_INSTALL_PREFIX`.
> Why not set `CPACK_PACKAGING_INSTALL_PREFIX` to `CMAKE_INSTALL_PREFIX` when DESTDIR is set?

See above.
> I am not sure I follow this. What do you mean "totally replaced"?

`dpkg` and `apt` report that the packages are totally replaced (presumably meaning that all dependencies and installed files are covered by other libraries, although specific documentation is difficult to find).
> But if its a header only then it should only be a dev package. I dont understand why it would provide a runtime package.

The "Provides" field on the dev package is set to the associated (nonexistant) runtime package, so that any previous packages or applications depending on the package (previously with the name now assigned to the runtime package) can still locate the correct files.

pfultz2:
> When component packaging is enabled, the TGZ generator uses both CPACK_PACKAGING_INSTALL_PREFIX and CMAKE_INSTALL_PREFIX, so an error is encountered with something like cannot lstat '/opt/rocm/opt/rocm' when CPACK_PACKAGING_INSTALL_PREFIX is set to CMAKE_INSTALL_PREFIX.

Ok got it, so an empty string makes sense.

> dpkg and apt report that the packages are totally replaced (presumably meaning that all dependencies and installed files are covered by other libraries, although specific documentation is difficult to find).

I still dont understand. Why does it think it can be replaced?

> The "Provides" field on the dev package is set to the associated (nonexistant) runtime package

Are you saying that we set the dev package's "Provides" field to the runtime package? That seems wrong. Why wouldn't we just make the dev package depend on the runtime package?

pfultz2:
> so that any previous packages or applications depending on the package (previously with the name now assigned to the runtime package) can still locate the correct files.

If an application depends on the header files at runtime then the dependencies should be updated to add the dev package.

saadrahim:
> 
> 
> > so that any previous packages or applications depending on the package (previously with the name now assigned to the runtime package) can still locate the correct files.
> 
> If an application depends on the header files at runtime then the dependencies should be updated to add the dev package.

This is to provide a transition mechanism. The provides will be removed, likely in 5.0, so that the provides value is not used afterwards. The only 3 libraries that use this are rocPRIM, rocThrust and hipCUB. There is no other impact.

lawruble13:
> I still dont understand. Why does it think it can be replaced?

The issue is the "runtime" packages for header-only libraries has no unique contents or dependencies, so because no files are actually installed when the "runtime" package would be installed, the package manager simply declines to install them.

lawruble13:
> Are you saying that we set the dev package's "Provides" field to the runtime package? That seems wrong. Why wouldn't we just make the dev package depend on the runtime package?

As mentioned by Saad, this is only used on the header-only libraries where the runtime package doesn't exist. All other libraries set the dev package to depend on the runtime package, and as a transition mechanism set the runtime package to have a weak dependency on the dev package (so any old code depending on a library at compile time still downloads what is expected).

pfultz2:
> The issue is the "runtime" packages for header-only libraries has no unique contents or dependencies, so because no files are actually installed when the "runtime" package would be installed, the package manager simply declines to install them.

So you are saying when we install a dev package(which implicitly depends on the runtime package) it will only install the dev package, and when it gets to the runtime package it ignores it since it has no files, which mean the dependencies needed for runtime package are also ignored(which the dependencies for the dev package is ignored). Is this what you mean?

So if we specify a HEADER_ONLY flag it should transfer the runtime dependencies to the dev package directly(ie something like `set(CPACK_DEBIAN_DEVEL_PACKAGE_DEPENDS ${CPACK_DEBIAN_PACKAGE_DEPENDS})`), but this needs to be done after `if(PARSE_DEPENDS)`.

lawruble13:
> So you are saying when we install a dev package(which implicitly depends on the runtime package) it will only install the dev package, and when it gets to the runtime package it ignores it since it has no files, which mean the dependencies needed for runtime package are also ignored(which the dependencies for the dev package is ignored). Is this what you mean?
> 
> So if we specify a HEADER_ONLY flag it should transfer the runtime dependencies to the dev package directly(ie something like `set(CPACK_DEBIAN_DEVEL_PACKAGE_DEPENDS ${CPACK_DEBIAN_PACKAGE_DEPENDS})`), but this needs to be done after `if(PARSE_DEPENDS)`.

The issue is that the dev package depends on the "runtime" package, so when the package manager attempts to install the dev package, it fails because it declines to install the "runtime" package, so the dev package has broken dependencies. 

The dependencies are correctly transferred to the devel package as-is, because `CPACK_DEBIAN_DEVEL_PACKAGE_DEPENDS` defaults to the value of `CPACK_DEBIAN_PACKAGE_DEPENDS` when `CPack` is included.

pfultz2:
> The issue is that the dev package depends on the "runtime" package, so when the package manager attempts to install the dev package, it fails because it declines to install the "runtime" package, so the dev package has broken dependencies.

So HEADER_ONLY we dont make the dev package depend on the runtime package?

lawruble13:
> > The issue is that the dev package depends on the "runtime" package, so when the package manager attempts to install the dev package, it fails because it declines to install the "runtime" package, so the dev package has broken dependencies.
> 
> So HEADER_ONLY we dont make the dev package depend on the runtime package?

Correct, if HEADER_ONLY is provided the dev package does not depend on the associated runtime package, and the dev package picks up any dependencies that were set for the package in general.

pfultz2:
> This is to provide a transition mechanism. The provides will be removed, likely in 5.0, so that the provides value is not used afterwards. The only 3 libraries that use this are rocPRIM, rocThrust and hipCUB. There is no other impact.

I see, so we probably shouldn't document that `HEADER_ONLY` adds a "Provides" field since this will be removed soon.

pfultz2:
It would be good to have a test for the `HEADER_ONLY` flag.

lawruble13:
> > This is to provide a transition mechanism. The provides will be removed, likely in 5.0, so that the provides value is not used afterwards. The only 3 libraries that use this are rocPRIM, rocThrust and hipCUB. There is no other impact.
> 
> I see, so we probably shouldn't document that `HEADER_ONLY` adds a "Provides" field since this will be removed soon.

Good point, although I think noting it would still be a good idea. I have updated the changelog to note that the "Provides" feature is introduced as deprecated, and will be removed in the future. Does that work for you?

pfultz2:
> Good point, although I think noting it would still be a good idea. I have updated the changelog to note that the "Provides" feature is introduced as deprecated, and will be removed in the future. Does that work for you?

Yea that works.

lawruble13:
> It would be good to have a test for the `HEADER_ONLY` flag.

I have added a test for a header only library, but I'm not sure in your testing framework how to check for the existence/properties of the generated package files.

pfultz2:
> I have added a test for a header only library, but I'm not sure in your testing framework how to check for the existence/properties of the generated package files.

There currently isnt, but at least the test can get some coverage.

PR #56, opened 2021-08-03
Rename debian DEVEL package to dev

lawruble13:
Debian packages are incorrectly named with -devel, they should be named with -dev.


saadrahim (approved):
This is to comply with our new package naming policies. Requirements were clarified to differential between development packages for debian and rpm. Debians will end in -dev and rpms in -devel.

PR #58, opened 2021-08-10
Client packaging functions

### LINKED ISSUE #55: Incorrect tests
https://github.com/RadeonOpenCompute/rocm-cmake/blob/48121ade0e70b89d393112642d8f394289fe680f/share/rocm/cmake/ROCMCreatePackage.cmake#L20

Using
           if(${inout_variable})
where the value of inout_variable might be CPACK_DEB_WIBBLE looks up the truthfulness of the variable CPACK_DEB_WIBBLE.
However the cmake idea of true does not include strings like
      N
      0
      FALSE
so this function givens incorrect results for

set(X FALSE)
rocm_join_if_set("," X "he said")

the function ought to be fixed to compare to the empty string or else renamed to rocm_join_if_true to reflect its nature. Of course the latter function is much less useful.

lawruble13:
Add functions to standardize how libraries package their clients (e.g. samples, tests, and benchmarks). 

Incidentally resolve #55 by correcting `rocm_join_if_set` and standardizing for external use.

Tests forthcoming.


CHANGELOG.md
```diff
@@ -1,5 +1,26 @@
 # Change Log for rocm-cmake
 
+## [0.7.0]
+### Added
+- Client packaging functionality:
+    - `rocm_setup_client_component` sets up a client component, with the given name and values for the component and associated package.
+    - `rocm_install_client_with_symlink` installs a single client target into a specified location, and installs a symlink to that target in a second specified location.
```
pfultz2, line 7: `rocm_install_targets` should be used to install targets so they are installed in the correct location. Whats the purpose of the symlinks?

share/rocm/cmake/ROCMClients.cmake
```diff
@@ -0,0 +1,81 @@
...
+
+macro(rocm_install_client_with_symlink TARGET COMPONENT)
+    set(options)
+    set(oneValueArgs EXE_DESTINATION SYMLINK_DESTINATION)
+    set(multipleValueArgs)
+
+    cmake_parse_arguments(PARSE "${options}" "${oneValueArgs}" "${multipleValueArgs}" ${ARGN})
+
+    if(NOT DEFINED PARSE_EXE_DESTINATION)
+        set(PARSE_EXE_DESTINATION "${CMAKE_PROJECT_NAME}/bin")
```
pfultz2, line 48: Exes should be installed in `${CMAKE_INSTALL_PREFIX}/bin`

README.md
```diff
@@ -133,3 +133,68 @@ rocm_check_target_ids
...
+rocm_add_deb_dependencies
+-------------------------
+
+    rocm_add_deb_dependencies(<deb-dep>...
+        [COMPONENT <component>]
+    )
+
+Adds the listed `<deb-dep>`s to the DEB package for the associated `<component>`, or for the package overall if a `<component>` is not given. NOTE: this dependency should be given in the RPM style (e.g. `foo >= 1.0`), NOT in the Debian style (e.g. `foo (>= 1.0)`).
+
+rocm_add_dependencies
```
pfultz2, line 193: I think these `rocm_add_*` functions should go under ROCMCreatePackage module. They should be named `rocm_package_add_*`. 

I assume these functions are required to be called before `rocm_create_package` so we should add a some kind of check that `rocm_create_package` hasn't been called.

README.md
```diff
@@ -133,3 +133,68 @@ rocm_check_target_ids
...
+---------------------
+
+    rocm_add_dependencies(<common-dep>...
```
pfultz2, line 196: The arguments should be `DEPENDS <common-dep>...`, this way we can add a `STATIC_DEPENDS` to specify additional dependencies when it is built statically.

pfultz2:
Also, not every library/package will create clients package, but it may create additional components. I think instead of `rocm_setup_client_component` we can have a `rocm_package_setup_component`:

```cmake
    rocm_package_setup_component(<component_name>
        [PACKAGE_NAME <package_name>]
        [LIBRARY_NAME <library_name>]
        [PARENT <parent-component>]
        [DEPENDS 
            [COMMON <common-dep>...]
            [DEBIAN <debian-dep>...]
            [RPM <rpm-dep>...]]
    )
```

Then client components can be created with:

```cmake
rocm_package_setup_component(clients)
# client1 component
rocm_package_setup_component(client1 PARENT clients)
# client2 component
rocm_package_setup_component(client2 PARENT clients)
```

When setting the parent it will then call `rocm_add_dependencies` for each client. Also, we could add a `COMPONENT` flag under `DEPENDS` so you can express dependencies on other components(without needing to figure it the components package name).

pfultz2:
You could still have a `rocm_setup_client_component` function to reduce the boilerplate in the libraries that have clients, but I think this function should build on top of something like `rocm_package_setup_component`, so we can easily add more components in the future.

share/rocm/cmake/ROCMClients.cmake
lawruble13: As these client exes are specific to each library, and are not generally useful programs, they are being installed to `${CMAKE_INSTALL_PREFIX}/<libname>/bin`, which is what is being done here (although the `${CMAKE_INSTALL_PREFIX}` is used implicitly, which may be worth revisiting, all other issues aside). They are then symlinked into `${CMAKE_INSTALL_PREFIX}/bin`, again using the implicit `${CMAKE_INSTALL_PREFIX}`.

README.md
lawruble13: I agree, and will rename and move these functions. For the check, I think that will print a warning but not cause an actual error, if that's acceptable? We may also want to rename these functions to better indicate that they don't add dependencies to an already created library, they only set up dependencies for when the package is created.

README.md
lawruble13: Good idea, will do that.

lawruble13:
Using a more general `rocm_package_setup_component` is a good idea, I'll start working on that. I do think having the `rocm_setup_client_component` function would still be useful, so I'll rework that to use the new function.

README.md
lawruble13: Should `STATIC_DEPENDS` specify additional dependencies, or alternative dependencies? If `STATIC_DEPENDS` would specify additional dependencies when built statically, should there be a `SHARED_DEPENDS` which specifies additional dependencies when built non-statically (i.e. `DEPENDS` would be the dependencies common to static and non-static builds, then each of those has its specific argument). For example, the math libraries clients depend on the base package when building shared libraries, but not when building static libraries.

pfultz2: > Should STATIC_DEPENDS specify additional dependencies, or alternative dependencies?

Currently, for [`rocm_export_targets`](https://github.com/RadeonOpenCompute/rocm-cmake#rocm_export_targets) it specifies additional dependencies. I dont think there will be a need for `SHARED_DEPENDS` as we have yet ran into an issue where the shared version has a dependency that the static build doesnt need. We can definitely add it in the future if that were to happen

README.md
pfultz2: > For the check, I think that will print a warning but not cause an actual error, if that's acceptable?

Thats fine for now, although I think it would be better for it to eventually become an error as it wont do what the user expects and developers shouldn't ignore it.

share/rocm/cmake/ROCMClients.cmake
pfultz2: [`rocm_install_targets`](https://github.com/RadeonOpenCompute/rocm-cmake#rocm_install_targets) already support this by setting the `PREFIX` flag to the subdir name(I guess its not well documented). And then [`rocm_install_symlink_subdir`](https://github.com/RadeonOpenCompute/rocm-cmake#rocm_install_symlink_subdir) can symlink everything in the subdir to `CMAKE_INSTALL_PREFIX`.

README.md
pfultz2: > We may also want to rename these functions to better indicate that they don't add dependencies to an already created library,

What do you mean "an already created library"? Do you mean the dependencies for cmake targets(ie `add_dependencies`)?

lawruble13: Sorry, that was poorly phrased. I meant we may want to name the functions something like `rocm_package_setup_*` rather than `rocm_package_add_*`, in order to indicate that the function should be called before calling `rocm_create_package`.

CHANGELOG.md
lawruble13: `rocm_install_targets` and `rocm_install_symlink_subdir` are now correctly used rather than `rocm_install_client_with_symlink`.

README.md
pfultz2: But can these functions be called multiple times to iteratively append the dependencies? If so, then `add` is better than `setup`.

lawruble13: Yes, they can, and that is a good point. The names have been left as-is.

README.md
lawruble13: As the client packages depend on the associated runtime packages only when the shared version is generated, I added both SHARED_DEPENDS and STATIC_DEPENDS to these functions.

share/rocm/cmake/ROCMInstallSymlinks.cmake
```diff
@@ -3,6 +3,11 @@
 # ######################################################################################################################
 
 function(rocm_install_symlink_subdir SUBDIR)
+    set(options)
+    set(oneValueArgs COMPONENT NAME_PREFIX)
```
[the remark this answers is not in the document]
lawruble13: It is a prefix that is added to the symlink, it is intended to help avoid name collisions (for example, the math libraries currently generate example executables named like `example_xyz`, where `xyz` is independent of the library and may be reused between libraries. Allowing a prefix to be added resolves this collision, and provides  information at a glance on which library installed a particular example.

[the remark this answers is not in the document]
pfultz2: Why not just set the `OUTPUT_NAME` property to the unique name? You can then add a `POST_BUILD` custom command to symlink(or copy) the short name in the build tree. You could add a function to take care of this.

It can be confusing to have two seperate names. A user script could fail if they were to add /opt/rocm/bin vs /opt/rocm/rocblas/bin to the path.

README.md
```diff
@@ -133,3 +189,46 @@ rocm_check_target_ids
     )
 
 Returns the subset of HIP [target-ids](https://clang.llvm.org/docs/ClangOffloadBundler.html#target-id) supported by the current CXX compiler.
+
+ROCMClients
+===========
+
+rocm_package_setup_client_component
```
pfultz2, line 196: I think it would be better to prefix the client functions with `rocm_client` which would make this function `rocm_client_package_setup_component`, but maybe thats not a better name.

share/rocm/cmake/ROCMCreatePackage.cmake
```diff
@@ -222,50 +272,44 @@ macro(rocm_create_package)
...
+                else()
+                    string(TOLOWER "${CPACK_PACKAGE_NAME}-${_upstream}" _upstream_rpm_package)
+                endif()
+                rocm_package_add_rpm_dependencies(COMPONENT "${_downstream}" DEPENDS "${_upstream_rpm_package} >= ${CPACK_PACKAGE_VERSION}")
+                rocm_package_add_deb_dependencies(COMPONENT "${_downstream}" DEPENDS "${_upstream_deb_package} >= ${CPACK_PACKAGE_VERSION}")
+            endforeach()
+        endif()
     endif()
     include(CPack)
+    set(ROCM_PACKAGE_CREATED TRUE CACHE INTERNAL "Track whether rocm_create_package has been called.")
```
pfultz2, line 300: I dont think a cached variable will work as this will be set to TRUE the second time cmake is ran. Instead. you could do `set(ROCM_PACKAGE_CREATED TRUE)` and then just check `if(DEFINED ROCM_PACKAGE_CREATED)` in the `rocm_package_add` functions.

share/rocm/cmake/ROCMCreatePackage.cmake
```diff
@@ -274,3 +318,45 @@ macro(rocm_set_comp_cpackvar HEADER_ONLY components)
...
+    if(DEFINED PARSE_PARENT)
+        list(APPEND ROCM_PACKAGE_COMPONENT_DEPENDENCIES "${PARSE_PARENT}->${COMPONENT_NAME}")
+    endif()
+
+    if(DEFINED PARSE_DEPENDS)
+        cmake_parse_arguments(DEPENDS "RUNTIME" "" "COMMON;DEBIAN;RPM;COMPONENT" ${PARSE_DEPENDS})
+        rocm_package_add_deb_dependencies(COMPONENT ${COMPONENT_NAME} DEPENDS ${DEPENDS_COMMON} ${DEPENDS_DEBIAN})
+        rocm_package_add_rpm_dependencies(COMPONENT ${COMPONENT_NAME} DEPENDS ${DEPENDS_COMMON} ${DEPENDS_RPM})
+        foreach(DEP_COMP IN LISTS DEPENDS_COMPONENT)
+            list(APPEND ROCM_PACKAGE_COMPONENT_DEPENDENCIES "${COMPONENT_NAME}->${DEP_COMP}")
```
pfultz2, line 356: So why couldn't you do `rocm_package_add_deb_dependencies(COMPONENT ${COMPONENT_NAME} DEPENDS ${CPACK_DEBIAN_${DEP_COMP}_PACKAGE_NAME})` to add the dependencies on other components?

share/rocm/cmake/ROCMCreatePackage.cmake
```diff
@@ -274,3 +318,45 @@ macro(rocm_set_comp_cpackvar HEADER_ONLY components)
...
+
+    set(CPACK_DEBIAN_${COMPONENT_GNAME}_PACKAGE_NAME "${PARSE_LIBRARY_NAME}-${PARSE_PACKAGE_NAME}")
+    set(CPACK_RPM_${COMPONENT_GNAME}_PACKAGE_NAME "${PARSE_LIBRARY_NAME}-${PARSE_PACKAGE_NAME}")
+
+    if(DEFINED PARSE_PARENT)
+        list(APPEND ROCM_PACKAGE_COMPONENT_DEPENDENCIES "${PARSE_PARENT}->${COMPONENT_NAME}")
+    endif()
+
+    if(DEFINED PARSE_DEPENDS)
+        cmake_parse_arguments(DEPENDS "RUNTIME" "" "COMMON;DEBIAN;RPM;COMPONENT" ${PARSE_DEPENDS})
```
pfultz2, line 352: Why cant `RUNTIME` be listed under `COMPONENT`?

share/rocm/cmake/ROCMCreatePackage.cmake
lawruble13: I used a cache variable to deal with the scoping constraints of CMake, because if we use `set(ROCM_PACKAGE_CREATED TRUE)` here then the variable will only be set in that scope. For example, in rocBLAS we perform a bunch of setup in the top level CMakeLists.txt, then include `cmake/packaging.cmake` to generate the package. Using the non-cache variable, if another component was set up in the top level CMakeLists.txt after including `cmake/packaging.cmake`, then that wouldn't be warned about. You are right though that a cache variable isn't ideal, as it does cause issues on a second run. I'll look into alternative solutions.

share/rocm/cmake/ROCMCreatePackage.cmake
lawruble13: The idea of doing it this way was that you can reference a component which hasn't been set up yet and still end up with the correct name, which was primarily to deal with the `Unspecified` (runtime) and `devel` components, which have their names set in `rocm_create_package`. Because this way the component dependencies aren't resolved to package dependencies until the very end of `rocm_create_package`, we can ensure that the dependencies use the same names as the generated packages.

lawruble13: We could also use this to ensure the `NAME` argument to `rocm_create_package` is properly propagated to the other components. Currently, unless the `LIBRARY_NAME` parameter is passed to `rocm_package_setup_component`, it just uses `PROJECT_NAME`, which is not always correct (e.g. when generating rocBLAS for Nvidia gpus, the project name is still `rocblas`, but we would like the package name to be `rocblas-alt`). Currently the components would be named with `rocblas`, but we could put in a placeholder when setting up the components, and finalize the names in the call to `rocm_create_package`.

share/rocm/cmake/ROCMCreatePackage.cmake
lawruble13: Another option would be to set `CMAKE_INSTALL_DEFAULT_COMPONENT_NAME` to `runtime` instead of `Unspecified`, then `rocm_package_setup_component([...] DEPENDS COMPONENT runtime [...])` would behave as expected.

README.md
lawruble13: I think that naming convention would also be fine, my thought process was primarily to name all of the new packaging related functions with `rocm_package`. Also, this function sets up a client component, whereas `rocm_package_setup_component` sets up a generic component, but the packaging is all the same, so having `client` closer to `component` makes slightly more sense to me.

All that said, I'm not particularly attached to this name, so if you think `rocm_client_package_setup_component` would be better I'm happy to change that.

share/rocm/cmake/ROCMCreatePackage.cmake
pfultz2: > Another option would be to set CMAKE_INSTALL_DEFAULT_COMPONENT_NAME to runtime instead of Unspecified

That would make things clearer.

share/rocm/cmake/ROCMCreatePackage.cmake
pfultz2: You could use a global property.

README.md
pfultz2: Yea, I am not sure if thats a better name, so if you dont think so either then lets leave it as it is.

share/rocm/cmake/ROCMCreatePackage.cmake
lawruble13: The solution I have currently implemented still uses a cache variable, but the variable is set to false at the top level of `ROCMCreatePackage.cmake`. This resets the variable to false on future builds, so there should be no spurious warnings. A global property would also work, but I think it would require slightly more boilerplate code.

share/rocm/cmake/ROCMCreatePackage.cmake
lawruble13: This has now been implemented.

share/rocm/cmake/ROCMCreatePackage.cmake
```diff
@@ -11,32 +13,103 @@ set(ROCM_DEP_ROCMCORE TRUE CACHE BOOL "Add dependency on rocm-core package, defa
...
+set(CMAKE_INSTALL_DEFAULT_COMPONENT_NAME runtime)
+set(ROCM_PACKAGE_CREATED FALSE CACHE INTERNAL "Track whether rocm_create_package has been called.")
+
+function(rocm_package_add_rpm_dependencies)
+    set(options QUIET)
+    set(oneValueArgs COMPONENT)
+    set(multiValueArgs DEPENDS SHARED_DEPENDS STATIC_DEPENDS)
+    cmake_parse_arguments(PARSE "${options}" "${oneValueArgs}" "${multiValueArgs}" ${ARGN})
+
+    if(${ROCM_PACKAGE_CREATED} AND NOT PARSE_QUIET)
```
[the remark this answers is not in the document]
lawruble13: The `QUIET` flag suppresses the error message about the function being called after the package was created. It was put in so that the error messages from `rocm_package_add_rpm_dependencies` and `rocm_package_add_deb_dependencies` could be suppressed in favor of a more helpful single message from `rocm_package_add_dependencies`. I then added it to `rocm_package_add_dependencies` for consistency.

PR #59, opened 2021-08-11
Add new implementation of rocm_join_if_set and tests

icarus-sparry:
Reimplement rocm_join_if_set so that it doesn't reject things that
cmake considers to match the boolean false, only empty strings

Add tests for the function.

Verify that both implementations behave the same as string(JOIN) with
respect to empty arguments.

Signed-off-by: Icarus Sparry <icarus.sparry@amd.com>


test/pass/rocm-list-split.cmake
```diff
@@ -1,4 +1,4 @@
-
+cmake_policy(SET CMP0011 NEW)
```
pfultz2, line 1: What is this for? Maybe this should be moved to the test.cmake file instead if its useful for all the tests.

icarus-sparry: This was to reduce the noise when running "make pass_rocm_list_plit" to run a single test.

PR #61, opened 2021-09-29
Use ROCm platform version to automatically set rocm-core dependency

lawruble13:
As the `rocm-core` package did not exist prior to ROCm version 4.5.0, packages built for prior versions of ROCm should not have this dependency. This code attempts to automatically determine the ROCm platform version, and if it is less than 4.5.0 then the package will not automatically depend on `rocm-core`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 # Change Log for rocm-cmake
 
+## [0.6.2]
+### Changed
+- If the ROCm platform version that is being built for is less than version 4.5.0, `ROCM_DEP_ROCMCORE` is automatically disabled.
+  - The ROCm platform version is determined by:
+    - the user-set CMake variable `ROCM_PLATFORM_VERSION`; otherwise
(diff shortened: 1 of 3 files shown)


saadrahim (approved):

CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
 # Change Log for rocm-cmake
 
+## [0.6.2]
+### Changed
+- If the ROCm platform version that is being built for is less than version 4.5.0, `ROCM_DEP_ROCMCORE` is automatically disabled.
+  - The ROCm platform version is determined by:
+    - the user-set CMake variable `ROCM_PLATFORM_VERSION`; otherwise
```
icarus-sparry, line 7: Just looking at the code, I only see commands to set ROCM_PLATFORM_VERSION, but without a test to see if it is already set. Am I missing something?

lawruble13: As `ROCM_PLATFORM_VERSION` is a cache variable, it is only set if there isn't an existing cache entry, so if the user has already set the value then the set commands do nothing.

pfultz2 (approved):

PR #63, opened 2022-01-13
Automatically install license

lawruble13:
If a license file was specified before calling rocm-create-package, install it to the correct location. If one wasn't specified, scan for a LICENSE(.*) file, then if one isn't found add it as the license file and install to the correct location.

### DIFF
--- a/share/rocm/cmake/ROCMCreatePackage.cmake
+++ b/share/rocm/cmake/ROCMCreatePackage.cmake
@@ -318,7 +318,7 @@ macro(rocm_set_comp_cpackvar HEADER_ONLY components)
     # Setting component specific variables
     set(CPACK_ARCHIVE_COMPONENT_INSTALL ON)
     if(NOT CPACK_RESOURCE_FILE_LICENSE)
-        file(GLOB _license_files LIST_DIRECTORIES FALSE "${CMAKE_SOURCE_DIR}/LICENSE.*")
+        file(GLOB _license_files LIST_DIRECTORIES FALSE "${CMAKE_SOURCE_DIR}/LICENSE*")
         list(LENGTH _license_files _num_licenses)
         if(_num_licenses GREATER 1)
             message(AUTHOR_WARNING "rocm-cmake warning: Multiple license files found, please specify one using CPACK_RESOURCE_FILE_LICENSE.")
--- a/share/rocm/cmake/ROCMCreatePackage.cmake
+++ b/share/rocm/cmake/ROCMCreatePackage.cmake
@@ -329,8 +329,12 @@ macro(rocm_setup_license HEADER_ONLY)
                 "rocm-cmake warning: Multiple license files found, "
                 "please specify one using CPACK_RESOURCE_FILE_LICENSE."
             )
-        elseif(_num_licenses GREATER 0)
-            message(STATUS "rocm-cmake: Set license file to ${CPACK_RESOURCE_FILE_LICENSE}.")
+        elseif(_num_licenses EQUAL 0)
+            message(AUTHOR_WARNING
+                "rocm-cmake warning: Could not find a license file, "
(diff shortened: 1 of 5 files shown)


cgmb (commented):
> If one wasn't specified, scan for a LICENSE(.*) file, then if one isn't found add it as the license file and install to the correct location.

I don't think this is a good idea. If there's some sort of edge case in which your scanner picks up the wrong file, it will probably take a while for anyone to notice. This is the sort of thing that developers should be explicit about.

saadrahim:
> > If one wasn't specified, scan for a LICENSE(.*) file, then if one isn't found add it as the license file and install to the correct location.
> 
> I don't think this is a good idea. If there's some sort of edge case in which your scanner picks up the wrong file, it will probably take a while for anyone to notice. This is the sort of thing that developers should be explicit about.

All libraries that are packaged now require a license file. I am not sure if there are edge cases. Perhaps limiting the scan to license.txt or license.md may be safer.

share/rocm/cmake/ROCMCreatePackage.cmake
```diff
@@ -317,6 +317,37 @@ endmacro()
 macro(rocm_set_comp_cpackvar HEADER_ONLY components)
     # Setting component specific variables
     set(CPACK_ARCHIVE_COMPONENT_INSTALL ON)
+    if(NOT CPACK_RESOURCE_FILE_LICENSE)
+        file(GLOB _license_files LIST_DIRECTORIES FALSE "${CMAKE_SOURCE_DIR}/LICENSE*")
```
saadrahim, line 321: There are 3 different license files names that we use. Should we limit it to just those the files? This may address @cgmb 's concerns

lawruble13: Seems like a reasonable solution to me, @cgmb? I will also note that it only looks for LICENSE files in the top-level CMake directory, which should help to ensure that only license files for the intended library are used.

share/rocm/cmake/ROCMCreatePackage.cmake
```diff
@@ -329,12 +329,8 @@ macro(rocm_setup_license HEADER_ONLY)
                 "rocm-cmake warning: Multiple license files found, "
                 "please specify one using CPACK_RESOURCE_FILE_LICENSE."
             )
-        elseif(_num_licenses EQUAL 0)
-            message(AUTHOR_WARNING
-                "rocm-cmake warning: Could not find a license file, "
-                "please specify one using CPACK_RESOURCE_FILE_LICENSE."
-            )
-        else()
```
pfultz2: We should keep this warning, and just add a LICENSE file to the tests.

lawruble13: Warning reinstated, and placeholder LICENSE files have been added to all tests which use rocm_create_package.

share/rocm/cmake/ROCMCreatePackage.cmake
cgmb: I was mostly worried about the use of `${CMAKE_SOURCE_DIR}`, but given that it's not recursive and you've added a check to ensure we find something, that's probably sufficient.

cgmb: That is, I think it's fine as is.

cgmb:
> I am not sure if there are edge cases.

I was thinking it was recursive, but `${CMAKE_SOURCE_DIR}` would be the source of most unexpected behaviour. Given that there's now a check that exactly one file is found, and Liam has reminded me that this isn't going to recurse through `CMAKE_BINARY_DIR` when that's a subdirectory of `CMAKE_SOURCE_DIR`, I think we're fine.

pfultz2 (approved):

PR #64, opened 2022-01-18
Set CPACK_RPM_PACKAGE_LICENSE

Mystro256:
Please let me know if this is the wrong place to set it, or if MIT is incorrect.

It's only set on the top level, not per component. If there are multiple licenses (such as different subpackages/components have different licenses), it should be separated by "AND" (mix of licenses) or "OR" (either license may be used) such as "MIT AND NCSA" or "MIT OR ASL 2.0". Some examples of licenses are here:
https://fedoraproject.org/wiki/Licensing:Main?rd=Licensing#Software_License_List

RH and SUSE generally copy Fedora's "short name" as it's intended to be very machine-readable.

### DIFF
--- a/share/rocm/cmake/ROCMCreatePackage.cmake
+++ b/share/rocm/cmake/ROCMCreatePackage.cmake
@@ -354,6 +354,11 @@ macro(rocm_setup_license HEADER_ONLY)
                 COMPONENT devel
             )
         endif()
+        file(READ ${CPACK_RESOURCE_FILE_LICENSE} LICENSEFILE)
+        string(FIND "${LICENSEFILE}" "MIT" MITCHECK)
+        if(NOT CPACK_RPM_PACKAGE_LICENSE AND ${MITCHECK} NOT EQUAL -1)
+                set(CPACK_RPM_PACKAGE_LICENSE "MIT")
+        endif()
     endif()
 endmacro()
 


Mystro256:
I realise this might be used for multiple components, are they all licensed MIT? I'm not really sure of the context of this function (e.g. is it used by more than just one packaged component?)

lawruble13:
This function is used for any component which uses rocm_create_package, which includes all of the `roc<lib>`s and `hip<lib>`s. While most libraries use an MIT license, rocThrust in particular uses Apache (as rocThrust inherits from a different project). As such, I don't think that fixing the MIT license will work.

share/rocm/cmake/ROCMCreatePackage.cmake
```diff
@@ -315,6 +315,8 @@ macro(rocm_create_package)
 endmacro()
 
 macro(rocm_setup_license HEADER_ONLY)
+    set(CPACK_RPM_PACKAGE_LICENSE "MIT")
```
lawruble13, line 318: This could set `CPACK_RPM_PACKAGE_LICENSE` to MIT by default, but it needs to be configurable per project. It should also be contained within the `if` block starting on line 345, so that if the license file is not detected or set, then no variables are set.

lawruble13 (changes requested):

share/rocm/cmake/ROCMCreatePackage.cmake
Mystro256: Does it make sense to set a default license string? I guess this pull request is more of throwing the idea out there.

I can wrap this in a "if CPACK_RPM_PACKAGE_LICENSE is not defined" if that makes sense to you.

lawruble13: Setting a default string would be convenient, but I could see it causing issues as it potentially obscures a mislabelling of the license in RPM packages (i.e. the license in the RPM would need to be checked against the license in the package to see issues where the variable wasn't correctly set, while with no default it's apparent because the license is 'unknown'). 

I'm torn though, as almost all projects use the MIT license, so having the default set makes a lot of sense and would save a lot of code duplication and PRs.

Mystro256:
Ok, see the updated patch, maybe this is a bit more sane.

If the license file is found automatically, see if the file contains a mention of "MIT", and if it does, set the RPM license string to MIT. I also updated the commit description to be less vague.

share/rocm/cmake/ROCMCreatePackage.cmake
```diff
@@ -354,6 +354,11 @@ macro(rocm_setup_license HEADER_ONLY)
                 COMPONENT devel
             )
         endif()
+        file(READ ${CPACK_RESOURCE_FILE_LICENSE} LICENSEFILE)
+        string(FIND "${LICENSEFILE}" "MIT" MITCHECK)
+        if(NOT CPACK_RPM_PACKAGE_LICENSE AND ${MITCHECK} NOT EQUAL -1)
+                set(CPACK_RPM_PACKAGE_LICENSE "MIT")
+        fi()
```
lawruble13, line 361: Closing if statements in CMake uses `endif`, not `fi`.

Mystro256: Thanks, I missed this. As per my above comment:

Do you have 2 components you suggest that I can test as an example? (one with MIT, one without)

lawruble13, line 361: Closing if statements in CMake uses `endif`, not `fi`.

lawruble13:
> Do you have 2 components you suggest that I can test as an example? (one with MIT, one without)

rocRAND uses an MIT license, rocThrust (as mentioned previously) uses the Apache license.

saadrahim:
What you are doing may have legal implications. I would be very cautious if there is a remote chance of misidentifying the license of a particular library or component. I do not think we should default to MIT even if most of our projects use that license.

share/rocm/cmake/ROCMCreatePackage.cmake
```diff
@@ -354,6 +354,11 @@ macro(rocm_setup_license HEADER_ONLY)
                 COMPONENT devel
             )
         endif()
+        file(READ ${CPACK_RESOURCE_FILE_LICENSE} LICENSEFILE)
+        string(FIND "${LICENSEFILE}" "MIT" MITCHECK)
```
cgmb, line 358: > THIS SOFTWARE IS PROVIDED ``AS IS″ AND WITHOUT ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, WITHOUT LI**MIT**ATION, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE.
?

Mystro256: Haha yeah good point

Mystro256:
I'm going to close this, I think #65 obsoletes this.

Thanks everyone for the input

lawruble13:
As @cgmb points out, this solution does find any instance of the string "MIT", including the unintended one he noted. One possible option would be to check for the string "MIT License" or something similar at position 0 in the string, which would only detect license files that specifically, as the first item in them, declare that they're MIT license files. That said, given the issues mentioned, I unfortunately find myself favouring specifying this variable on a per-project basis.

PR #65, opened 2022-01-18
Adding parameter for License type and License file in rocm_create_pac…

nunnikri:
Adding parameter for License type and License file in rocm_create_package

### DIFF
--- a/share/rocm/cmake/ROCMCreatePackage.cmake
+++ b/share/rocm/cmake/ROCMCreatePackage.cmake
@@ -134,7 +134,7 @@ endmacro()
 
 macro(rocm_create_package)
     set(options LDCONFIG PTH HEADER_ONLY)
-    set(oneValueArgs NAME DESCRIPTION SECTION MAINTAINER LDCONFIG_DIR PREFIX)
+    set(oneValueArgs NAME DESCRIPTION SECTION MAINTAINER LDCONFIG_DIR PREFIX LICENSE LICENSE_PATH)
     set(multiValueArgs DEPENDS COMPONENTS)
 
     cmake_parse_arguments(PARSE "${options}" "${oneValueArgs}" "${multiValueArgs}" ${ARGN})
@@ -159,6 +159,15 @@ macro(rocm_create_package)
     rocm_get_patch_version(ROCM_VERSION_NUM)
     if(ROCM_VERSION_NUM)
         set(CPACK_PACKAGE_VERSION "${CPACK_PACKAGE_VERSION}.${ROCM_VERSION_NUM}")
+    endif()
+    # Set the License to rpm header
+    if(PARSE_LICENSE)
+        set( CPACK_RPM_PACKAGE_LICENSE "${PARSE_LICENSE}" )
+    endif()
+    # Copy the license file to /opt/rocm/share/doc/<component_name>
+    if(PARSE_LICENSE_PATH)
+        set(CPACK_RESOURCE_FILE_LICENSE "${PARSE_LICENSE_PATH}")
+        install(FILES ${CPACK_RESOURCE_FILE_LICENSE} DESTINATION share/doc/${PARSE_COMPONENT})
     endif()
 
     set(CPACK_DEBIAN_PACKAGE_MAINTAINER ${PARSE_MAINTAINER})


share/rocm/cmake/ROCMCreatePackage.cmake
```diff
@@ -160,6 +160,15 @@ macro(rocm_create_package)
...
         set(CPACK_PACKAGE_VERSION "${CPACK_PACKAGE_VERSION}.${ROCM_VERSION_NUM}")
     endif()
+    # Set the License to rpm header
+    if(PARSE_LICENSE)
+        set( CPACK_RPM_PACKAGE_LICENSE "${PARSE_LICENSE}" )
+    endif()
+    # Copy the license file to /opt/rocm/share/doc/<component_name>
+    if(PARSE_LICENSE_PATH)
+        set(CPACK_RESOURCE_FILE_LICENSE "${PARSE_LICENSE_PATH}")
+        install(FILES ${CPACK_RESOURCE_FILE_LICENSE} DESTINATION share/doc/${PARSE_COMPONENT})
```
lawruble13, line 170: `PARSE_COMPONENT` will not be defined.

lawruble13: The install statement is also redundant (see line 344) and not entirely correct, as it installs the license only to the `runtime` component, while header only libraries need this file installed to the `devel` component.

Mystro256: For reference, the desired location is:

share/doc/XXX

where XXX is either the cmake project name OR the package name.

E.g. share/doc/rocrand for "rocRAND"

nunnikri: @lawruble13  thanks for the comments. PARSE_COMPONENT was a copy paste mistake. Correcting it to CPACK_PACKAGE_NAME. 
If devel package is depended on main package we need the license file only in 1 package. Like Jeremy pointed out rocrand and rocrand-dev package licence will be in share/doc/rocRAND

lawruble13: @nunnikri For header only libraries, there is no runtime package (as there would be no contents to that package), so the license file must be installed to the devel package. For all other libraries, the runtime component doesn't depend on the devel component, so the runtime component must be the component which has the license file.

lawruble13: The installation is already handled on the aforementioned lines, simply removing the install statement on line 170 should work.

nunnikri: Thanks Liam. The changes are updated as per your comments

share/rocm/cmake/ROCMCreatePackage.cmake
```diff
@@ -160,6 +160,15 @@ macro(rocm_create_package)
...
         set(CPACK_PACKAGE_VERSION "${CPACK_PACKAGE_VERSION}.${ROCM_VERSION_NUM}")
     endif()
+    # Set the License to rpm header
+    if(PARSE_LICENSE)
+        set( CPACK_RPM_PACKAGE_LICENSE "${PARSE_LICENSE}" )
+    endif()
+    # Setting the CPACK_RESOURCE_FILE_LICENSE variable.
+    # Installing license file to /opt/rocm/share/doc/<component_name> will happen in rocm_setup_license macro
+    if(PARSE_LICENSE_PATH)
+        set(CPACK_RESOURCE_FILE_LICENSE "${PARSE_LICENSE_PATH}")
```
Mystro256, line 170: CPACK_RESOURCE_FILE_LICENSE might not be a requirement.

I think it's only used by Windows CMake, but I'm not sure.

lawruble13: It is still automatically detected and set (if possible) in `rocm_setup_license`.

saadrahim (changes requested):
Please review https://github.com/RadeonOpenCompute/rocm-cmake/pull/63 

I am not sure if this PR is needed in its current form.

lawruble13:
I agree with Saad, adding these parameters doesn't actually accomplish much. The same purpose could be accomplished by setting the appropriate CPack variables in the relevant library prior to calling `rocm_setup_package`, and as this PR adds no further automation it doesn't cut down on the additions to other libraries.

nunnikri:
@lawruble13 @saadrahim @pfultz2  we can do all the CPACK variable setting outside of create_package fn. But I think its always better to encapsulate inside the fn whatever possible. If all of you thinks the change is still not required, let me know I can close the PR

share/rocm/cmake/ROCMCreatePackage.cmake
```diff
@@ -160,6 +160,15 @@ macro(rocm_create_package)
...
+    endif()
```
pfultz2, line 171: We should add an install statement to install the license file.

lawruble13: The installation is done in the `rocm_setup_license` macro added in #63.

share/rocm/cmake/ROCMCreatePackage.cmake
```diff
@@ -160,6 +160,15 @@ macro(rocm_create_package)
     if(ROCM_VERSION_NUM)
         set(CPACK_PACKAGE_VERSION "${CPACK_PACKAGE_VERSION}.${ROCM_VERSION_NUM}")
     endif()
+    # Set the License to rpm header
+    if(PARSE_LICENSE)
+        set( CPACK_RPM_PACKAGE_LICENSE "${PARSE_LICENSE}" )
+    endif()
+    # Setting the CPACK_RESOURCE_FILE_LICENSE variable.
+    # Installing license file to /opt/rocm/share/doc/<component_name> will happen in rocm_setup_license macro
```
pfultz2, line 168: Where is this `rocm_setup_license` at? I dont see any documentation on it.

share/rocm/cmake/ROCMCreatePackage.cmake
pfultz2: Users are supposed to call that? There is no documentation for it.

share/rocm/cmake/ROCMCreatePackage.cmake
lawruble13: It is located on line 326, it does still need to be documented. Apologies on that, this feature was requested to be enabled ASAP and documentation was overlooked.

share/rocm/cmake/ROCMCreatePackage.cmake
lawruble13: It's not intended to be called by users, it's called in `rocm_set_comp_cpackvar`.

share/rocm/cmake/ROCMCreatePackage.cmake
pfultz2: Ok, looking closer this is called only when `COMPONENTS` are passed. Also, if users are supposed to call it, then the interface should be changed to use named paramaters instead of passing a bool.

lawruble13: Opened #67 to resolve issues with `rocm_setup_license`, as they are unrelated to this PR.

PR #68, opened 2022-01-20
Rearranged Postinst and prerm used for Python setting

raramakr:
Added a function call for python Library path setting 
Function will be called in single Rocm setup
In Multi Rocm setup the function call will be stripped from package
User has to provide the library path to python via environmental variable in Multi ROCM setup


share/rocm/cmake/ROCMCreatePackage.cmake
```diff
@@ -292,20 +292,31 @@ macro(rocm_create_package)
...
-            ")
-
-            file(
-                APPEND ${PROJECT_BINARY_DIR}/debian/prerm
-                "
-                rm ${PYTHON_SITE}/${PARSE_NAME}.pth
-            ")
-        endforeach()
+        list(GET PYTHON_SITE_PACKAGES 0 PYTHON_SITE_0)
+        list(GET PYTHON_SITE_PACKAGES 1 PYTHON_SITE_1)
```
pfultz2, line 296: This wont work if `PYTHON_SITE_PACKAGES` has only 1 or 3 paths. This is a cached variable so it can be overwritten by the user to have any number paths. This should still use `foreach` loop to generate the pre/post scripts.

pfultz2 (changes requested):

share/rocm/cmake/ROCMCreatePackage.cmake
```diff
@@ -292,20 +292,31 @@ macro(rocm_create_package)
...
+            APPEND ${PROJECT_BINARY_DIR}/debian/postinst
+            "
+            set_libdir(){
+                mkdir -p ${PYTHON_SITE_0}
+                echo \"${LIB_DIR}\" > ${PYTHON_SITE_0}/${PARSE_NAME}.pth
+
+                mkdir -p ${PYTHON_SITE_1}
+                echo \"${LIB_DIR}\" > ${PYTHON_SITE_1}/${PARSE_NAME}.pth
+            }
+            set_libdir
```
pfultz2, line 307: Shouldn't ldconf be added into this function also?

share/rocm/cmake/ROCMCreatePackage.cmake
raramakr: Thanks Paul .Will change as per the review comment

share/rocm/cmake/ROCMCreatePackage.cmake
frepaul: Since only a handful of packages use PYTHON path, I would suggest to keep in separate.
From what I remember ldconfig enablement is controlled by a different variable as well.

share/rocm/cmake/ROCMCreatePackage.cmake
raramakr: Changes are made . Pls review

raramakr: Changes are made . Pls review

pfultz2 (approved):

share/rocm/cmake/ROCMCreatePackage.cmake
```diff
@@ -132,6 +132,46 @@ macro(rocm_package_add_rocm_core_dependency)
     endif()
 endmacro()
 
+function(parse_python_syspath)
```
pfultz2, line 135: This function should be prefixed with `rocm_`.

pfultz2 (changes requested):

share/rocm/cmake/ROCMCreatePackage.cmake
raramakr: sure , will add prefix rocm_

saadrahim (changes requested):
Do we need documentation and changelog updates? Do we need to update the version number?

share/rocm/cmake/ROCMCreatePackage.cmake
```diff
@@ -132,6 +132,46 @@ macro(rocm_package_add_rocm_core_dependency)
...
+    file(APPEND ${PROJECT_BINARY_DIR}/debian/prerm
+        "
+        rm_libdir(){
+    ")
+    foreach(PYTHON_SITE ${PYTHON_SITE_PACKAGES})
+        file(
+            APPEND ${PROJECT_BINARY_DIR}/debian/postinst
+            "
+            mkdir -p ${PYTHON_SITE}
+            echo \"${LIB_DIR}\" > ${PYTHON_SITE}/${PARSE_NAME}.pth
```
lawruble13, line 153: I'm wary of using `PARSE_NAME` here, if the `rocm_create_package` variables ever change their names this would be very easy to miss as it's in a different scope. Consider passing this in as a parameter?

raramakr: Thanks @lawruble13 for the comments. Will work on that

share/rocm/cmake/ROCMCreatePackage.cmake
pfultz2: > From what I remember ldconfig enablement is controlled by a different variable as well.

A "different" variable? There is no variable to disable the PTH files, only for ldconf. Since its a cmake variable, it requires to a different different build, right?

From what I understand is that you are just removing the `set_libdir` line from the packaging scripts for multi-rocm builds. We could do the same for ldconf, and we dont need two different builds(it also seems simpler).

I guess we can address this in a future PR though.

pfultz2:
> Do we need documentation and changelog updates? Do we need to update the version number?

There could be an update to the changelog, but I dont see a need to change the documentation(this is essentially doing the same thing from a user perspective).

cgmb:
> There could be an update to the changelog, but I dont see a need to change the documentation(this is essentially doing the same thing from a user perspective).

That is true, but the `PTH` option appears to be undocumented at the moment.

share/rocm/cmake/ROCMCreatePackage.cmake
raramakr: Passed PARSE_NAME and LIB_DIR as function arguments

pfultz2 (approved):

PR #70, opened 2022-01-25
Create better default CPACK_GENERATOR

icarus-sparry:
Currently if a package doesn't specify a CPACK_GENERATOR the generated
one is excessive. rocBLAS generates several tarballs which are about
1GB in size each which are not used. This consumes I/O time slowing
the builds as well as using disk. Systems using "deb" files need to
install programs to create "rpm" packages and vice versa. With
generation of debug information these requirements become even more
stringent, for example the standard macros for packaging rpm debug
info require eu-strip to be installed.

If the build environment specifies a PKGTYPE variable and no
CPACK_GENERATOR is specified then set CPACK_GENERATOR based on this
value.

Change-Id: I14dcf59f6681d985f8e9ca25e68b4a3537835800

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Change Log for rocm-cmake
 
+## [0.7.2]
+- `rocm_create_package` now will attempt to set a default `CPACK_GENERATOR` based on the `PKGTYPE` environment variable if one is not already set.
+
 ## [0.7.1]
 ### Added
 - `CLANG_TIDY_USE_COLOR` variable added to control the color output from `clang-tidy`, by default this is `On`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Change Log for rocm-cmake
 
 ## [0.7.2]
-- `rocm_create_package` now will attempt to set a default `CPACK_GENERATOR` based on the `PKGTYPE` environment variable if one is not already set.
+- `rocm_create_package` now will attempt to set a default `CPACK_GENERATOR` based on the `ROCM_PKGTYPE` environment variable if one is not already set.
 
 ## [0.7.1]
 ### Added
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,7 +17,7 @@ list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_LIST_DIR}/share/rocm/cmake)
 include(ROCMCreatePackage)
 include(ROCMSetupVersion)
 
-rocm_setup_version(VERSION 0.7.1)
+rocm_setup_version(VERSION 0.7.2)
 
 include(CMakePackageConfigHelpers)
 write_basic_package_version_file(${CMAKE_CURRENT_BINARY_DIR}/ROCMConfigVersion.cmake
--- a/doc/src/ROCMCreatePackage.rst
+++ b/doc/src/ROCMCreatePackage.rst
@@ -22,7 +22,7 @@ Commands
 Sets up CPack packaging, including installing the license file to the correct location and component.
 If the license file is not specified, also attempts to locate a LICENSE, LICENSE.md, or LICENSE.txt file in `CMAKE_SOURCE_DIR`.
 
-If the CPACK_GENERATOR has not been specified then if PKGTYPE is set
+If the CPACK_GENERATOR has not been specified then if ROCM_PKGTYPE is set
 then use that to set it, otherwise probe the system to see what
 programs are available.
 
--- a/doc/src/ROCMCreatePackage.rst
+++ b/doc/src/ROCMCreatePackage.rst
@@ -22,6 +22,10 @@ Commands
 Sets up CPack packaging, including installing the license file to the correct location and component.
 If the license file is not specified, also attempts to locate a LICENSE, LICENSE.md, or LICENSE.txt file in `CMAKE_SOURCE_DIR`.
 
+If the CPACK_GENERATOR has not been specified then if PKGTYPE is set
+then use that to set it, otherwise probe the system to see what
+programs are available.
+
 .. cmake:command:: rocm_package_add_rpm_dependencies
 
 .. code-block:: cmake
--- a/share/rocm/cmake/ROCMCreatePackage.cmake
+++ b/share/rocm/cmake/ROCMCreatePackage.cmake
@@ -172,6 +172,53 @@ function(rocm_parse_python_syspath DIR_PATH PKG_NAME)
     ")
 endfunction()
 
+macro(rocm_set_cpack_gen)
+    # If CPACK_GENERATOR value has been given, then just use it
+    if(NOT CPACK_GENERATOR)
+	# If there is a PKGTYPE, use that as the desired type
+	if(DEFINED ENV{PKGTYPE})
+	    set(CPACK_GENERATOR "" ) # Create the variable if needed
+	    string(TOUPPER $ENV{PKGTYPE} CPACK_GENERATOR) # PKGTYPE is typically lower case
+	else()
+	    # Otherwise see what we can find
+	    message(INFO "rcom_set_cpack_gen didn't find PKGTYPE in environment")
+	    set(CPACK_GENERATOR "TGZ;ZIP")
+	    if(EXISTS ${MAKE_NSIS_EXE})
+		list(APPEND CPACK_GENERATOR "NSIS")
+	    endif()
+
+	    if(EXISTS ${RPMBUILD_EXE})
+		list(APPEND CPACK_GENERATOR "RPM")
+	    endif()
+
+	    if(EXISTS ${DPKG_EXE})
+		list(APPEND CPACK_GENERATOR "DEB")
+	    endif()
+	endif()
+    endif()
+    # Set up some additional variables depending on which generator we are going to use
+    if (CPACK_GENERATOR MATCHES ".*RPM.*")
+	if(PARSE_COMPONENTS)
+	    set(CPACK_RPM_COMPONENT_INSTALL ON)
+	endif()
+    endif()
+    if (CPACK_GENERATOR MATCHES ".*DEB.*")
+	if(EXISTS ${DPKG_EXE})
+	    if(PARSE_COMPONENTS)
+		set(CPACK_DEB_COMPONENT_INSTALL ON)
+		execute_process(
+		    COMMAND dpkg --print-architecture
+		    RESULT_VARIABLE PROC_RESULT
+		    OUTPUT_VARIABLE COMMAND_OUTPUT
+		    OUTPUT_STRIP_TRAILING_WHITESPACE)
+		if(PROC_RESULT EQUAL "0" AND NOT COMMAND_OUTPUT STREQUAL "")
+		    set(CPACK_DEBIAN_PACKAGE_ARCHITECTURE "${COMMAND_OUTPUT}")
+		endif()
+	    endif()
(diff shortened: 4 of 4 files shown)


pfultz2:
Why not set CPACK_GENERATOR to the  packages you want when you call cmake?

icarus-sparry:
"Why not set CPACK_GENERATOR to the packages you want when you call cmake?" the code is just about setting a better default.

If CPACK_GENERATOR is set, then it is honored. If it is not set, but PKGTYPE is set then that is used, otherwise use the existing code.

If the calls to cmake were in a central place, where we could add flags to set the packaging type with ease then we wouldn't need to do this. As it is most (all) of the projects which consume rocm-cmake have their own "install" or "install.sh" scripts to invoke cmake, with different conventions (do I use '-k' to set CMAKE_BUILD_TYPE to RelWithDebInfo, or something else or does the script not allow this to be set at all?). Setting the type of packaging desired is something that should be a choice of the build infrastructure, not the project developer.

frepaul:
> Why not set CPACK_GENERATOR to the packages you want when you call cmake?

Seems like the code will take care of this. macro checks for  CPACK_GENERATOR is not SET.  This should make the code better if user fail to provide right value. Since some of the packaging step particularly for debug-symbols packaging do not work well without additional tools across OS, this should help to make it work with default set of tools in respective OS. Also this help to speed up the build on our CI that has different build profiles for each OS by avoiding unnecessary packaging step and throwing away that package!

pfultz2:
> If the calls to cmake were in a central place, where we could add flags to set the packaging type with ease then we wouldn't need to do this.

I see, I thought there was a cmake toolchain that was passed to all projects.

> If CPACK_GENERATOR is set, then it is honored. If it is not set, but PKGTYPE is set then that is used, otherwise use the existing code.

If we are going to provide a env variable workaround for our install.sh scripts, I think it should at least be named `CPACK_GENERATOR` to make clearer what its setting in cmake.

> As it is most (all) of the projects which consume rocm-cmake have their own "install" or "install.sh" scripts to invoke cmake, with different conventions (do I use '-k' to set CMAKE_BUILD_TYPE to RelWithDebInfo, or something else or does the script not allow this to be set at all?). 

Well not all projects. MIGraphX and MIOpen doesnt, although it uses rbuild, you can pass cmake flags and toolchain files directly to rbuild, or just use rbuild for dependencies and then invoke cmake directly.

However, these custom install.sh scripts should be gotten rid of since you can't pass cmake flags nor toolchain file nor invoke cmake directly.

pfultz2:
> Seems like the code will take care of this. macro checks for CPACK_GENERATOR is not SET.

Thats not really my question. The question is why do we want to set a `PKGTYPE` env variable instead of setting the `CPACK_GENERATOR` variable?

cgmb:
> However, these custom install.sh scripts should be gotten rid of since you can't pass cmake flags nor toolchain file nor invoke cmake directly.

I agree with this. Many external users skip the install.sh script and invoke CMake directly. It's fine to have a helper install.sh script for component developers to put their commonly used options in a shared place, but the important interface is actually the direct CMake interface. IMO, invoking CMake directly should be the canonical way to build each library, and it's how the libraries should be built and tested by our tooling. (I'm not sure every stakeholder agrees with me, but that's what I think.)

icarus-sparry:
> If we are going to provide a env variable workaround for our install.sh scripts, I think it should at least be named `CPACK_GENERATOR` to make clearer what its setting in cmake.

The point is to provide a reasonable default. The CI needs to know the packaging type for other purposes, for example todecide which files to archive. You can argue for a different name, but I think that tying the general purpose name to CPACK is wrong.  I suggest if we already need a general purpose variable then we should use it. There are discussions about dropping cpack due to its conformance ( e.g. the packages it produces have non standard names) and performance ( e.g. it doesn't use multiple cores for compression) problems.  I would not be adverse to adding CPACK_GENERATOR as an additional environment variable, although I see this as adding an unneeded extra choice. If we do add CPACK_GENERATOR then the logic should be to use it in preference to PKGTYPE if both are specified.

pfultz2:
The problem with `PKGTYPE` it just seems arbitrary and doesn't have `ROCM_` prefix to avoid name conflicts. Since the goal is to set the `CPACK_GENERATOR` I would prefer the env variable to be named the same. 

I dont see how dropping cpack is relevant here.

icarus-sparry:
@pfultz2 The CI already sets PKGTYPE, which is why I proposed it here. Yes of course it is arbitary. Will you approve this PR if I change the code to use ROCM_PKGTYPE and change the CI to match?

icarus-sparry:
There are 2 major reasons for using functions/macros in high level code. One is to reduce typing and errors. The more import one is [information hiding](https://en.wikipedia.org/wiki/Information_hiding). It enables you to have a function like **rocm_create_package** whose name tells you what it is going to do, but not how it is going to do it. This allows you to change the implementation if you want, which is a very good thing.

Unfortunately the [original version](https://github.com/RadeonOpenCompute/rocm-cmake/commit/11bcc88221a938e8755b8cbfc3f53a3be8744d7f#diff-c3734e9136634b1fab1490000459155afb7b3d0362fa32a4741fe5f560aec757R6)  of this didn't set up passing a parameter to specify what kind of package was required. Instead it unconditionally decided to create **deb** files and if it found a **makensis** program it would create a windows installer as well. These choices were not something the user could alter. [Later](https://github.com/RadeonOpenCompute/rocm-cmake/commit/3cd270c0dfc3c71b3dcc2cc5a80361204960133d#diff-c3734e9136634b1fab1490000459155afb7b3d0362fa32a4741fe5f560aec757) code was added to create **rpm** files if a **rpmbuild** program was found, and after that code was [added](https://github.com/RadeonOpenCompute/rocm-cmake/commit/77c14e5851efc0a44cdef4de62659f75c65ffb34#diff-c3734e9136634b1fab1490000459155afb7b3d0362fa32a4741fe5f560aec757) to always create **zip** and **tar** and only create **deb** files if **dpkg** was found.

As there was no parameter for saying what kind of package was requested, the CMakeLists.txt files which called this had no incentive to worry about what kind of packages were being generated, they just took what they were given. Likewise the scripts which called cmake had no incentive to request only particular types of packages. The build environment could of course ensure there were no dpkg or rpmbuild to influence things but this is an excessive need, and doesn't address tarball creating which is handled internally by cpack.

The design, where you look for programs in the build environment and based on that decide what kind of packages to produce makes it seductively easy to start using, it all appears to "just work". However it is fragile. In particular if I have a program **rpmbuild** on my Ubuntu machine then it will produce **rpm** files even if I don't want them. Worse still is if the **rpmbuild** program doesn't work, then the build will fail. 

The intent of this patch is to allow an environment variable that will allow the build environment to say what kind of packages are wanted, so a Ubuntu build environment can say that it wants just **deb** files and not **tar**, **rpm** or **nsis** files. Likewise a CENTOS build environment can ask for just **rpm** files even if there is a **dpkg** program in it.

If the callers of **rocm_create_package** have explicitly asked for some particular type then that is honored, otherwise give the build environment a chance to say what it can provide, otherwise fall back to the existing code.

The reason for not calling the environment variable **CPACK_GENERATOR** is for information hiding, just as the **rocm_create_package** wasn't called **rocm_create_package_using_cpack** to hide the information that it is using cpack. This is particularly relevant as we are considering changing away from cpack.  If we do change to use different tools for packaging then the hope is that we can just rewrite **rocm_create_package** to use these different tools. Probably at that stage we would add additional parameters to **rocm_create_package** to specify the type(s) of packages desire, but it still makes sense for the build environment to provide an environment variable to provide a good default. Having it called **CPACK_GENERATOR** when we are not using cpack and wouldn't be using it to set the **CPACK_GENERATOR** cmake variable would be confusing. We could start out with this name now and then rename it at that time but we  wouldn't want to have to rewrite all the CMakeLists.txt files which use the function and the scripts which call cmake to handle this internal implementation change.

saadrahim:
Should we link this to the OS detection capabilities in rocm-cmake and set the package type appropriately based on OS?

cgmb:
> The reason for not calling the environment variable CPACK_GENERATOR is for information hiding, just as the rocm_create_package wasn't called rocm_create_package_using_cpack to hide the information that it is using cpack. This is particularly relevant as we are considering changing away from cpack. If we do change to use different tools for packaging then the hope is that we can just rewrite rocm_create_package to use these different tools. Probably at that stage we would add additional parameters to rocm_create_package to specify the type(s) of packages desire, but it still makes sense for the build environment to provide an environment variable to provide a good default. Having it called CPACK_GENERATOR when we are not using cpack and wouldn't be using it to set the CPACK_GENERATOR cmake variable would be confusing. We could start out with this name now and then rename it at that time but we wouldn't want to have to rewrite all the CMakeLists.txt files which use the function and the scripts which call cmake to handle this internal implementation change.

While I largely agree with up until this point, I don't think this is a very strong argument. The libraries that use rocm-cmake already have places where they interact with cpack. Switching to a different system is going to take more than just swapping out the implementation of `rocm_create_package`.

Personally, I like your `ROCM_PKGTYPE` suggestion more than `CPACK_PKGTYPE` because it makes it very clear to me that the variable is something that we have invented, and that I should be looking through our code to learn more (not the cpack documentation).

icarus-sparry:
> Should we link this to the OS detection capabilities in rocm-cmake and set the package type appropriately based on OS?

If I understand the question then I think the answer is "no". I see these as only loosely related. The idea of cmake and related utilities is to provide a cross-platform way of building. Cpack can build a deb file without external tools so could build one on Windows. It can also build a tar file. On the other hand it can't build an rpm without the rpmbuild program.

I think the list of package types to be built is something that the user should set, not the author of the rocBLAS CMakeLists.txt file and not the author of the rocm-cmake ROCMCreatePackage.cmake file although both of them are welcome to provide defaults. If a user wants a tarball then they should get one. If they don't want one then we can save time and disk space by not creating one.

I don't see a way to determine based on the OS to decide if a user wants a tarball. Likewise if the user wants a deb package created or not.

icarus-sparry:
> While I largely agree with up until this point, I don't think this is a very strong argument. The libraries that use rocm-cmake already have places where they interact with cpack. Switching to a different system is going to take more than just swapping out the implementation of `rocm_create_package`.

I agree. I like to imagine I will get to live in an environment where everything is well designed, future requirements are anticipated and there are no bugs. Originally this macro ended with the include of cpack, so it was easier to imagine that all that was needed was to swap out the implementation of rocm_create_package, but those times have passed. The open question is if we can recapture them?

pfultz2:
> Should we link this to the OS detection capabilities in rocm-cmake and set the package type appropriately based on OS?

No, since we could be cross-compiling.

cgmb (commented):
I'm fine with this, but I don't want to approve anything without @pfultz2. 

One other suggestion that came up was forwarding `-D` arguments through from install.sh to CMake. That seems like something that would be nice to have on all the math libraries.

cgmb:
A different, but related question:

How do you folks feel about defaulting [`CPACK_DEBIAN_COMPRESSION_TYPE`](https://cmake.org/cmake/help/latest/cpack_gen/deb.html#variable:CPACK_DEBIAN_COMPRESSION_TYPE) to xz? For large libraries, our DEB packages are ~2.5x the size of our RPMs because the former defaults to gzip and the latter to xz. The installation of xz packages has been supported since Ubuntu 14.04, so I think the only drawback is the time to generate the package, which can be mitigated with multi-threaded compression controlled by [`CPACK_THREADS`](https://cmake.org/cmake/help/latest/module/CPack.html#variable:CPACK_THREADS)

PR #72, opened 2022-02-09
CPACK_GENERATOR fixes

lawruble13:
The `rocm_set_cpack_gen` call needs to be moved to the original location of the code dealing with `CPACK_GENERATOR`, as some of the body of `rocm_create_package` modifies the `PARSE_COMPONENTS` variable. The current location does not see these modifications, and so does not generate components with them.

### DIFF
--- a/share/rocm/cmake/ROCMCreatePackage.cmake
+++ b/share/rocm/cmake/ROCMCreatePackage.cmake
@@ -181,7 +181,7 @@ macro(rocm_set_cpack_gen)
             string(TOUPPER $ENV{ROCM_PKGTYPE} CPACK_GENERATOR) # PKGTYPE is typically lower case
         else()
             # Otherwise see what we can find
-            message(INFO "rcom_set_cpack_gen didn't find PKGTYPE in environment")
+            message(INFO "rocm_set_cpack_gen didn't find PKGTYPE in environment")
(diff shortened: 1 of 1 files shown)


share/rocm/cmake/ROCMCreatePackage.cmake
```diff
@@ -181,7 +181,7 @@ macro(rocm_set_cpack_gen)
             string(TOUPPER $ENV{ROCM_PKGTYPE} CPACK_GENERATOR) # PKGTYPE is typically lower case
         else()
             # Otherwise see what we can find
-            message(INFO "rcom_set_cpack_gen didn't find PKGTYPE in environment")
+            message(INFO "rocm_set_cpack_gen didn't find PKGTYPE in environment")
```
cgmb, line 184: Should it also say `ROCM_PKGTYPE` rather than `PKGTYPE`?

cgmb (approved):
Ideally, every bugfix would come with a test that fails before the fix and passes after. That said, I understand if there isn't infrastructure for that at the moment.

share/rocm/cmake/ROCMCreatePackage.cmake
lawruble13: Yes, it should. Updated.

PR #75, opened 2022-02-26
Make rocm_check_target_ids more robust

Maetveis:
Add support for target features for rocm_check_target ids. Before this change targets with target features specified (like xnack or sramecc) would fail. 

EDIT: the autodetection only fails (at least on clang) when `-Werror` is included in `CMAKE_CXX_FLAGS` because by default clang only issues a diagnostic not an error.

### DIFF
--- a/share/rocm/cmake/ROCMCheckTargetIds.cmake
+++ b/share/rocm/cmake/ROCMCheckTargetIds.cmake
@@ -30,12 +30,12 @@ function(rocm_check_target_ids VARIABLE)
 endfunction()
 
 function(_rocm_sanitize_target_id TARGET_ID VARIABLE)
-    # Cmake defines a preprocessor macro with this value, so it must be a valid C identifier
+    # CMake defines a preprocessor macro with this value, so it must be a valid C identifier
     # Handle + and - for xnack and sramecc so that e.g. xnack+ and xnack- doesn't get folded to
     # the same string by MAKE_C_IDENTIFIER
     string(REPLACE "_" "__"   TARGET_ID "${TARGET_ID}")
-    string(REPLACE "+" "_ON"  TARGET_ID "${TARGET_ID}")
-    string(REPLACE "-" "_OFF" TARGET_ID "${TARGET_ID}")
+    string(REPLACE "+" "_on"  TARGET_ID "${TARGET_ID}")
+    string(REPLACE "-" "_off" TARGET_ID "${TARGET_ID}")
     string(MAKE_C_IDENTIFIER "${TARGET_ID}" TARGET_ID)
     set(${VARIABLE} "${TARGET_ID}" PARENT_SCOPE)
 endfunction()
--- a/share/rocm/cmake/ROCMCheckTargetIds.cmake
+++ b/share/rocm/cmake/ROCMCheckTargetIds.cmake
@@ -20,7 +20,7 @@ function(rocm_check_target_ids VARIABLE)
 
     foreach(_target_id ${PARSE_TARGETS})
         _rocm_sanitize_target_id("${_target_id}" _result_var)
-        set(_result_var "HAVE_${_result_var}")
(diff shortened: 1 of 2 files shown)


share/rocm/cmake/ROCMCheckTargetIds.cmake
```diff
@@ -1,5 +1,5 @@
 # ######################################################################################################################
-# Copyright (C) 2021 Advanced Micro Devices, Inc.
+# Copyright (C) 2022 Advanced Micro Devices, Inc.
```
saadrahim, line 2: 2021-2022

share/rocm/cmake/ROCMCheckTargetIds.cmake
```diff
@@ -19,11 +19,23 @@ function(rocm_check_target_ids VARIABLE)
...
         check_cxx_compiler_flag("-xhip --offload-arch=${_target_id}" "${_result_var}")
         if(${_result_var})
             list(APPEND _supported_target_ids "${_target_id}")
         endif()
     endforeach()
-    set(${VARIABLE} "${_supported_target_ids}" PARENT_SCOPE)
+    set("${VARIABLE}" "${_supported_target_ids}" PARENT_SCOPE)
+endfunction()
+
+function(_rocm_sanitize_target_id TARGET_ID VARIABLE)
```
saadrahim, line 32: Can you link to a code snippet that will use this function? Do we need a test for this function in rocm-cmake?

saadrahim (changes requested):

share/rocm/cmake/ROCMCheckTargetIds.cmake
saadrahim: I mean a link to a problem you are facing in a problem outside of rocm-cmake. I want to understand more about the problem you are facing.

Maetveis: I actually found this with rocPRIM, but it does not actually cause issues for us. Since I was curious I looked at why its failing and made a simple fix. Here's a simple repoducer: https://github.com/Maetveis/rocm-cmake-repro.

The second function is only for internal use in rocm-cmake thats why its name starts with an underscore. Regarding tests I didn't find any in the code for rocm_check_target_ids, but I can add one for this case. The only issue with that is the detection needs hipcc or hip-clang and currently it is not part of this repro's CI.

cgmb:
Thanks for pointing out this problem. When I wrote the function, I was actually a little surprised that it allowed the unmodified target-id names. Your PR looks reasonable, but I'll try reproducing the problem and test your changes.

If we're changing the result variable anyway, it might be a good opportunity to improve the name in other ways. Whenever I see `-- Performing Test HAVE_gfx90a:xnack-` I have a tinge of regret that I didn't name it something more clear. Maybe something like `COMPILER_HAS_TARGET_ID_<sanitized-target-id>`.

share/rocm/cmake/ROCMCheckTargetIds.cmake
saadrahim: Hi @lawruble13, should we add CI for rocm-cmake to a CI system that has hipcc available? No certain if it worth the extra effort but I feel more confident if all code paths are tested.

pfultz2: We dont need hipcc just clang installed. We can setup dockers with github actions to do this.

cgmb:
I've reproduced the problem and tested before and after with rocBLAS. These are some nice changes. I'd be happy to approve this.

I made a couple tweaks to the variable name and I think I'm satisfied. This presumably needs a changelog entry, though I'm not sure exactly what heading it should be under. I noticed we've been bumping the version on a lot of changes recently. Is there a criteria we're using for that, @lawruble13?

cgmb (approved):

saadrahim (approved):

PR #77, opened 2022-03-09
Correct typo in rocm_wrap_header_dir

lawruble13:


cgmb:
Every bugfix should come with a test that fails before the change and passes after. The fact that this slipped through is an indicator that our current tests are insufficient.

lawruble13:
@cgmb I enhanced the existing test for the wrapper to cover this scenario, in addition to one other which was fixed in #76.

test/libwrapper/wrapper.cpp
```diff
@@ -5,3 +5,9 @@
 
 void wrapper()
 {}
+
+void other()
+{}
+
+void tld()
+{}
```
cgmb, line 13: As of C++11, it's technically valid for a source file to end without a newline, but it's good practice anyway. If you changed the file extension to `.c`, it would be undefined behaviour even in C17 (as per §5.1.1.2).

cgmb (approved):
LGTM. I had a minor suggestion.

test/libwrapper/wrapper.cpp
lawruble13: Good point, and good suggestion, thanks.

PR #79, opened 2022-03-18
Correct exports when using components

lawruble13:
`rocm_install_targets` automatically adds all targets to the export file. When the targets may be going into another component, this can cause issues (see SWDEV-328320). This PR changes the export code so that each component has an associated `-targets` file, and changes the configuration script to check for any of these files and include/set the appropriate values.


pfultz2:
Exactly how does this work? If the user wants to use the targets from a different component do they call `find_package(<some-other-component-name>)` or `find_package(<name> COMPONENT <component-name>)` or how does a user do `find_package` to get these components?

Usually the export file is installed in a different component than the target itself(ie runtime vs dev package). It seems this would get confusing when tying the component name to the export file, but I guess I dont fully understand how this works.

There is already an `EXPORT` argument that can be passed to use a differently named export, however, there is no argument to change the config.cmake name used, but this could be added.

lawruble13:
As is, if the user wanted to use the targets from another component they would still call `find_package(<name>)`, and would need to check for the target's validity. We could use the COMPONENT keyword, which I think would be a good idea. 

The way that it works is that if the user calls `rocm_export_targets` with `COMPONENT`, it creates a new target file, and sets up the config file to include that target file if it's present.

The problem it attempts to solve is that if you have multiple components with targets, the single current target file will fail unless all of the components are installed (e.g. with the client packages, which contain executables, installing just runtime and devel won't work because of the exported targets in the other packages, as in the linked issue). Just changing the name won't be sufficient, as the call to `rocm_export_targets` overwrites the package config file and only accepts the one target file. Doing this with `find_package(... COMPONENTS ...)` would be cleaner, as there would still only be the one file which adds targets depending on the requested components.

cgmb:
I will have to do some studying before I can give meaningful input on this. The export and install process is, unfortunately, the part of the build that I'm least familiar with. While I more or less understand how it's done in a simple C++ CMake project, I haven't dived into how rocm-cmake changes, improves, or facilitates that.

I suppose my first question is: why do we want to export targets for test executables in our installed client packages?

lawruble13:
Long and short of it is we don't, but I didn't want to preclude targets in any other component from potentially being exported in the future (for instance, installing asan libraries to an asan package)

cgmb:
> Long and short of it is we don't, but I didn't want to preclude targets in any other component from potentially being exported in the future (for instance, installing asan libraries to an asan package)

I'd rather we cross that bridge when we come to it. As it stands, we're paying the price of complexity today for a feature that we _might_ want some time in the future. I appreciate the work you did on this, but it seems like it would be simpler and safer just to stop exporting targets for the client components.

I'm not really opposed to adding this feature, but I don't fully understand the implications and that makes me hesitant.

PR #80, opened 2022-03-21
Always create runtime package if not using devel component

lawruble13:
Header-only libraries don't have a runtime component, unless the devel component isn't being used (e.g. on Windows). If the devel component isn't being used, then the runtime component should always be packaged.


cgmb (approved):
I suppose this fixes the rocRAND package? LGTM.

lawruble13:
rocRAND is correct, AFAIK? It fixes rocPRIM, rocThrust, and hipCUB.

PR #81, opened 2022-03-22
Don't export executable targets

lawruble13:
Alternative to #79, so that executable targets are not exported.

### DIFF
--- a/share/rocm/cmake/ROCMInstallTargets.cmake
+++ b/share/rocm/cmake/ROCMInstallTargets.cmake
@@ -127,22 +127,26 @@ function(rocm_install_targets)
             PATTERN "*.inl")
     endforeach()
 
-    install(
-        TARGETS ${PARSE_TARGETS}
-        EXPORT ${EXPORT_FILE}
-        RUNTIME
-            DESTINATION ${BIN_INSTALL_DIR}
-            COMPONENT ${runtime}
-        LIBRARY
-            DESTINATION ${LIB_INSTALL_DIR}
-            COMPONENT ${runtime}
-            NAMELINK_SKIP
-        ARCHIVE
-            DESTINATION ${LIB_INSTALL_DIR}
-            COMPONENT ${development}
-    )
     foreach(TARGET IN LISTS PARSE_TARGETS)
         get_target_property(T_TYPE ${TARGET} TYPE)
+        set(export_arg EXPORT ${EXPORT_FILE})
+        if(T_TYPE STREQUAL "EXECUTABLE")
+            unset(export_arg)
+        endif()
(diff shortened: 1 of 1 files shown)


lawruble13:
@cgmb @pfultz2 This is causing issues with getting the client packaging into mainline, so if we can get this in without too much delay that'd be appreciated.

saadrahim (approved):

cgmb (commented):
I can live with this, assuming that we weren't actually using this feature to export executable targets in any previous release. It would just need a changelog update.

I would like to know what @pfultz2 thinks, though.

lawruble13:
I'm obviously not enthralled with this solution, but the other simple option would be to not export any targets which are going to a component other than runtime, which seemed much more likely to cause issues. To the best of my knowledge, the math libraries weren't exporting any executables, so the question is whether any of the MI libraries did so, so I'd also like to hear from @pfultz2 before merging this.

pfultz2:
> This is causing issues with getting the client packaging into mainline, so if we can get this in without too much delay that'd be appreciated.

You could just set a different export file(via the `EXPORT` parameter) for the client's targets(and just ignore it) so there is no need to update rocm-cmake to fix mainline. This will give us time to design this feature right as its a little confusing what the `COMPONENT` parameter applies to. 

> I'm obviously not enthralled with this solution, but the other simple option would be to not export any targets which are going to a component other than runtime, which seemed much more likely to cause issues.

We could add a `NO_EXPORT` flag to `rocm_install` to skip exporting targets. So if developers want to export the targets they can add it as a dependency to the dev component otherwise use `NO_EXPORT`. We could also add a way to create different config.cmake for a different `EXPORT` file and component(or make it work with `find_package(... COMPONENT ...)`.

Maybe there is a way to track such dependencies between components, so we could show a warning when `NO_EXPORT` or `EXPORT` is not passed to targets in a component that doesn't depend on the component that installs such export file, or maybe implicitly provide the dependency while warning as well.

> To the best of my knowledge, the math libraries weren't exporting any executables

So the missing targets from the client package are coming from libraries not executables?

pfultz2:
Ah wait, I see this is a different PR to only export non-executables. If this fixes the issue in mainline then we can merge it. Sorry for the noise.

pfultz2 (approved):

cgmb (approved):
That's certainly not noise. Those are good suggestions. They sound a lot more like a proper fix.

lawruble13:
Concur with Cory, those are good suggestions Paul and we should incorporate those in the future. However as a stopgap measure to fix mainline for 5.2 FC, I'm going to merge this in.

PR #82, opened 2022-03-25
Create project-specific link to lib folder

lawruble13:
Adds the global option `ROCM_SYMLINK_LIBS`, which will create a symlink '/opt/rocm/<project-name>/lib` to `../lib`. This is for backwards compatibility, as most components do not use `rocm_install_symlink_subdir` anymore.


share/rocm/cmake/ROCMInstallTargets.cmake
```diff
@@ -157,6 +159,24 @@ function(rocm_install_targets)
                         NAMELINK_ONLY
             )
         endif()
+        if(T_TYPE MATCHES ".*_LIBRARY" AND ROCM_SYMLINK_LIBS AND NOT ROCM_CREATED_${PROJECT_NAME}_SYMLINK)
+            set(ROCM_CREATED_${PROJECT_NAME}_SYMLINK TRUE CACHE INTERNAL "Record symlink created.")
+            string(TOLOWER "${PROJECT_NAME}" LINK_SUBDIR)
+            if(CMAKE_HOST_WIN32)
+                set(SYMLINK_CMD "file(COPY \${LINK_TGT} DESTINATION \${LINK_NAME})")
+            else()
```
TorreZuk, line 167: I would suggest we go directly to new structure on windows as don't have significant downstream users and for windows .dll the single path addition is much simpler.

share/rocm/cmake/ROCMInstallTargets.cmake
```diff
@@ -157,6 +159,24 @@ function(rocm_install_targets)
...
+                set(SYMLINK_CMD "execute_process(COMMAND ln -sf ../lib \${LINK_NAME})")
```
TorreZuk, line 168: Can test later but this LINK_NAME is just lib for rocblas right?   So you are in the project subfolder already.

lawruble13: No, it's a bit more complex due to needing to create the symlink in the package, this method of doing it creates this symlink in the correct location at install time, which may be a more complex location (e.g. when CPack is creating the packages, the link is created directly inside CPack's internal staging directory).

TorreZuk:
Should have Centos and SLES test pipelines too.

lawruble13:
This is used across all libraries on all OSes, testing rocm-cmake on Centos and SLES is not super useful as the behaviour on all Linux systems is almost identical. The libraries should be tested on other OSes as well, I have a build that compiles all libraries against this PR going here: http://math-ci.rocm.amd.com/blue/organizations/jenkins/compute-rocm-dkms-no-npi-hipclang%2Fmeta%2Funified-1%2Fpull-requests/detail/pull-requests/880/pipeline

TorreZuk (approved):
Seems to me to do what we require.  Can anyone else think of potential bad sideeffects?

share/rocm/cmake/ROCMInstallTargets.cmake
```diff
@@ -157,6 +159,19 @@ function(rocm_install_targets)
                         NAMELINK_ONLY
             )
         endif()
+        if(T_TYPE MATCHES ".*_LIBRARY" AND ROCM_SYMLINK_LIBS AND NOT ROCM_CREATED_${PROJECT_NAME}_SYMLINK
```
pfultz2, line 162: Is `ROCM_CREATED_${PROJECT_NAME}_SYMLINK` variable to only insert the symlink once?

share/rocm/cmake/ROCMInstallTargets.cmake
```diff
@@ -157,6 +159,19 @@ function(rocm_install_targets)
...
+            AND NOT CMAKE_HOST_WIN32)
+
+            set(ROCM_CREATED_${PROJECT_NAME}_SYMLINK TRUE CACHE INTERNAL "Record symlink created.")
+            string(TOLOWER "${PROJECT_NAME}" LINK_SUBDIR)
```
pfultz2, line 166: This does assume that the project name and subdir name are the same. I believe this is so on most projects, but is that the case for all projects we have?

pfultz2:
> Adds the global option ROCM_SYMLINK_LIBS, which will create a symlink 

Hopefully we will be switching this off in the future.

pfultz2 (approved):

share/rocm/cmake/ROCMInstallTargets.cmake
lawruble13: Yes. It shouldn't have any side effects if it were to be inserted multiple times, but I wanted to guard against it anywya.

share/rocm/cmake/ROCMInstallTargets.cmake
lawruble13: For all of the math libraries, this is the case, I can't speak to any others.

lawruble13:
> Hopefully we will be switching this off in the future.

Absolutely concur, I'd be happy to see this switched off or removed, it's just necessary to maintain backwards compatibility while we change things.

PR #84, opened 2022-03-31
Only link library specific files

lawruble13:
This is not compatible with CMake 3.10 (due to the use of generator expressions in `install(CODE)`, requiring the setting of `CMP0087`). Can anyone think of a better method to do the same thing?

### DIFF
--- a/share/rocm/cmake/ROCMInstallTargets.cmake
+++ b/share/rocm/cmake/ROCMInstallTargets.cmake
@@ -346,5 +346,4 @@ function(rocm_export_targets)
             ")
         rocm_install(SCRIPT "${CMAKE_CURRENT_BINARY_DIR}/cmake_symlink.cmake")
     endif()
-    
 endfunction()
--- a/share/rocm/cmake/ROCMInstallTargets.cmake
+++ b/share/rocm/cmake/ROCMInstallTargets.cmake
@@ -159,7 +159,8 @@ function(rocm_install_targets)
                         NAMELINK_ONLY
             )
         endif()
-        if(ROCM_SYMLINK_LIBS AND NOT WIN32 AND T_TYPE MATCHES ".*_LIBRARY")
+        if(ROCM_SYMLINK_LIBS AND NOT WIN32 AND T_TYPE MATCHES ".*_LIBRARY"
+            AND NOT T_TYPE STREQUAL "INTERFACE_LIBRARY")
             string(TOLOWER "${PROJECT_NAME}" LINK_SUBDIR)
             file(GENERATE OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/${TARGET}_symlink.cmake
                 CONTENT "
--- a/share/rocm/cmake/ROCMInstallTargets.cmake
+++ b/share/rocm/cmake/ROCMInstallTargets.cmake
@@ -159,26 +159,24 @@ function(rocm_install_targets)
                         NAMELINK_ONLY
             )
         endif()
-        if(T_TYPE MATCHES ".*_LIBRARY" AND ROCM_SYMLINK_LIBS AND NOT CMAKE_HOST_WIN32)
-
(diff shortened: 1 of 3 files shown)


share/rocm/cmake/ROCMInstallTargets.cmake
```diff
@@ -159,16 +159,25 @@ function(rocm_install_targets)
                         NAMELINK_ONLY
             )
         endif()
-        if(T_TYPE MATCHES ".*_LIBRARY" AND ROCM_SYMLINK_LIBS AND NOT ROCM_CREATED_${PROJECT_NAME}_SYMLINK
-            AND NOT CMAKE_HOST_WIN32)
+        if(T_TYPE MATCHES ".*_LIBRARY" AND ROCM_SYMLINK_LIBS AND NOT CMAKE_HOST_WIN32)
+            cmake_policy(SET CMP0087 NEW)
```
cgmb, line 163: The use of [`CMAKE_HOST_WIN32`](https://cmake.org/cmake/help/latest/variable/CMAKE_HOST_WIN32.html) rather than [`WIN32`](https://cmake.org/cmake/help/latest/variable/WIN32.html) surprises me. `CMAKE_HOST_WIN32` describes the platform that you're running on and `WIN32` describes the platform that you're building for. I'd expect the latter to be what you care about in this code.

(They'll be identical unless you're cross-compiling, so the distinction is not usually important. Still, it's worth using the variable with the semantics that match your intent.)

cgmb (commented):
Perhaps you can get the name from [`get_target_property`](https://cmake.org/cmake/help/latest/command/get_target_property.html)? I'm not sure which it corresponds to, but I imagine it can be found somewhere in the list of [target properties](https://cmake.org/cmake/help/latest/manual/cmake-properties.7.html#properties-on-targets).

share/rocm/cmake/ROCMInstallTargets.cmake
lawruble13: This was copied from `rocm_install_symlink_subdir`, I believe due to the use of `ln` as opposed to `cmake -E create_symlink`.

cgmb:
If using the temporary script works, I have no real objections to that approach.

cgmb (approved):

TorreZuk:
Why are we running macOS tests?  @pfultz2 @saadrahim  should they block Liam

lawruble13:
I updated the noted line, however CMake seems to be unable to locate `cl` on the new version.

lawruble13:
CMake 3.10 is not aware of and does not support Visual Studio 2019. The earliest version which does is [CMake 3.14](https://cmake.org/cmake/help/v3.14/variable/MSVC_VERSION.html#variable:MSVC_VERSION). @pfultz2 preferred resolution?

pfultz2:
We should then skip 3.10 on windows. There is a way to exclude combinations described here:

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-excluding-configurations-from-a-matrix

There is also an `if` that can be added to skip steps:

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsif

pfultz2:
Let me adjust the required statuses so this can be merged.

lawruble13:
Any idea why the macos tests aren't running?

TorreZuk:
> Any idea why the macos tests aren't running?

No clue on the test queue, but I would at least vote to make the macos tests not required, are you ready to merge otherwise?    I don't even support running them until we actually have use cases on macos.    This fix is blocking downstream builds so hope it can get in today if you are happy with it so we can iterate if it causes some new issues.  @DanyiLin has a windows build pipeline consuming rocm-cmake so will be interesting to see if this will trigger it.

cgmb:
Perhaps try bumping it to macOS-10.15? I'm not certain where to find the supported platforms list for GitHub Actions, but macOS 10.14 isn't in [Azure's list of Microsoft-hosted agents](https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops&tabs=yaml).

PR #85, opened 2022-04-05
Remove INFOrocm_set_cpack_gen message

cgmb:
INFO is not a valid mode for CMake [`message`](https://cmake.org/cmake/help/latest/command/message.html), so it is interpreted as a plain string, printing "INFOrocm_set_cpack_gen didn't..." The message is not particularly useful anyway, so it might as well be removed.


pfultz2 (approved):
Yes, it bothered me this useless spam.

lawruble13 (approved):
I'm happy with this once the tests are resolved.

PR #87, opened 2022-04-13
Use CMake default locations for wrappers

lawruble13:
When `CMAKE_INSTALL_PREFIX` and `CPACK_PACKAGING_INSTALL_PREFIX` are both empty/unset, use the default for `CMAKE_INSTALL_PREFIX` as the installation prefix. This prevents the `file(ABSOLUTE_PATH)` statements from erroring out without an install prefix specified, and also alters the "correct" header file location in the generated wrapper. The relative paths are unaffected.


cgmb (approved):
I do sort of wonder if specifying the absolute path in the warning was a mistake. Nevertheless, this is a small, incremental fix for a corner case... and I definitely prefer those sorts of fixes at this point.

lawruble13:
Unless there are any objections to this, I'll merge it in later today after I finish validating it against the math libraries.

PR #88, opened 2022-04-17
Set default CPACK_DEBIAN_COMPRESSION_TYPE to xz

cgmb:
Comparison of rocBLAS 5.1.1 Debian package build options on a 3950X workstation:

```
 Time | Compression Type | Size
 -----|------------------|-----
  26s |       gzip       | 222M
 184s |  xz (1 thread)   |  61M
  19s |  xz (16 threads) |  63M
  ```

The original defaults can be restored by building with the CMake
variables `CPACK_DEBIAN_COMPRESSION_TYPE=gzip` and `CPACK_THREADS=1`
or the environment variables `ROCM_DEBIAN_COMPRESSION_TYPE=gzip` and
`ROCM_PKGTHREADS=1`.

### CODE REVIEW

File: share/rocm/cmake/ROCMCreatePackage.cmake [hunk 1]
```diff
@@ -218,6 +218,30 @@ macro(rocm_set_cpack_gen)
...
+    endif()
+
+    # xz compression is slow, but can be parallelized with CMake 3.20 or later.
+    # The compression ratio is marginally affected. The speedup is subject to
+    # diminishing returns at higher thread counts.
+    if(NOT CPACK_THREADS)
+        if(DEFINED ENV{ROCM_PKGTHREADS})
+            set(CPACK_THREADS $ENV{ROCM_PKGTHREADS})
+        else()
+            set(CPACK_THREADS "8")
```
pfultz2, line 240: Should we use `ProcessorCount` to set the number of threads?

share/rocm/cmake/ROCMCreatePackage.cmake
```diff
@@ -218,6 +218,30 @@ macro(rocm_set_cpack_gen)
...
+            set(CPACK_DEBIAN_COMPRESSION_TYPE $ENV{ROCM_DEBIAN_COMPRESSION_TYPE})
+        else()
+            set(CPACK_DEBIAN_COMPRESSION_TYPE "xz")
+        endif()
+    endif()
+
+    # xz compression is slow, but can be parallelized with CMake 3.20 or later.
+    # The compression ratio is marginally affected. The speedup is subject to
+    # diminishing returns at higher thread counts.
+    if(NOT CPACK_THREADS)
```
pfultz2, line 236: Rather than doing `if(NOT CPACK_THREADS)` can we just make these cached variables instead?

share/rocm/cmake/ROCMCreatePackage.cmake
cgmb: That would be `set(CPACK_THREADS "0")`. I also think that would be a better default, but I hadn't tested on a 128-core machine and wanted to pick something I knew was safe.

I'm not certain, but I think the concurrency is bounded by `(total size) / (block size)` anyway, because going beyond 8 threads made no noticeable difference in my tests on rocBLAS and rocSOLVER.

In any case, I'll stop being lazy and I'll test out "0" on a server with lots of cores. It's probably fine.

cgmb: The rocSOLVER package built in 12 seconds.

share/rocm/cmake/ROCMCreatePackage.cmake
cgmb: I initially agreed with you, but I took a look at how CMake handles this sort of thing for the `CMAKE_CXX_COMPILER_LAUNCHER` cache and environment variable, since that has a similar behaviour. [They use exactly the same sort of check](https://github.com/Kitware/CMake/blob/v3.23.1/Modules/CMakeCXXInformation.cmake#L219-L222), despite it being seemingly redundant.

I think the reasoning is probably that they don't want the variable to show up as an editable cache entry if the library build code is overriding the cache entry with a normal variable. It would be kind of annoying to edit the cache entry only to discover that it doesn't have any effect.

So, instead I made them cache variables in addition to doing `if(NOT CPACK_THREADS)`.

cgmb: I asked in the cmake channel of the cpplang slack about why CMake did it that way and Alex Reinking pointed me to the bug report that lead to CMP0126: [Set-cache deletes normal variables unpredictably (#22038)](https://gitlab.kitware.com/cmake/cmake/-/issues/22038).

lawruble13 (approved):
This looks good to me, I'll be glad to see these packages get smaller.

pfultz2 (approved):

PR #90, opened 2022-04-19
Add ROCM_NO_WRAPPER_HEADER_WARNING

cgmb:
This is checks if `ROCM_NO_WRAPPER_HEADER_WARNING` is defined and applies the same logic as was already being used for `ROCM_@ITEM_GUARD@_GAVE_WARNING`. It basically exists so users can use a more meaningful define in their build code than `ROCM_SYMLINK_GAVE_WARNING` and `ROCM_WRAPPER_GAVE_WARNING`. The template header diff might look a bit messy because I inverted the if/else.

The `pass/wrapper-nowarn.cmake` tests were copied from `fail/wrapper.cmake` but had `-DROCM_NO_WRAPPER_HEADER_WARNING` added to their compiler flags. There's also minor fixes to the docs and I changed the comment style to be C89 compatible... just in case.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [0.7.3]
 ### Added
 - Header wrapper functionality included. This is to support the change in header file locations in ROCm 5.2, while providing backwards compatibility via header file wrappers.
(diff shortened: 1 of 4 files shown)


lawruble13 (approved):
LGTM, thanks for providing the pretty option.

PR #91, opened 2022-04-21
Add original file contents to wrapper

lawruble13:
Tensorflow and other frameworks occasionally read a header file directly, looking for a specific string. While the location has been updated in the latest versions, earlier versions still look in the incorrect location. This code allows the contents of those header files to be added to the wrapper file, without affecting any program compiled using the wrapper.


cgmb:
It's worth understanding [the Tensorflow build code that is broken by the wrapper headers](https://github.com/tensorflow/tensorflow/blob/cd8dc5a6e87a0b21b8aa8f5a8f5179edbbe50120/third_party/gpus/find_rocm_config.py#L300-L328).

The method Tensorflow is using to detect the version information is very fragile. Many of these headers are implementation details, and are not designed to be directly included by user code (let alone parsed with a regex). We can add a workaround for the problem in the short term if need be, but I hope that for the long term we can find a better way for Tensorflow to get the information it needs.

I'm not familiar enough with the Bazel build system to know if there's a better way to detect the library version. It's been a very long time since I tangled with Bazel, but the starting place is probably [bazel.build: Working with External Dependencies](https://bazel.build/docs/external#non-bazel-projects).

nunnikri:
We tested the change in few of the math lib components. It seems to be working. 
Shall we merge it.

cgmb (commented):
I have a couple minor comments, but mostly this ~~needs the CI errors addressed and~~ the merge conflict resolved.

EDIT: You got the CI errors already. :+1:

cgmb (approved):

PR #92, opened 2022-04-22
Set Windows to use devel component

lawruble13:


saadrahim (approved):

saadrahim:
@DanyiLin just letting you know, we will have two packages soon. One a develop package and one a runtime package.

cgmb (approved):
This looks fine, assuming you've done some testing and found that it works as expected. We should be certain that @DanyiLin is ready for it before merging, though.

DanyiLin:
> This looks fine, assuming you've done some testing and found that it works as expected. We should be certain that @DanyiLin is ready for it before merging, though.

It works for windows CI

Issue #86, opened 2022-04-08
circular symlink issue: link <package>/lib/lib<package>.so to /opt/rocm/lib/lib<package>.so


yhmtsai:
https://github.com/RadeonOpenCompute/rocm-cmake/blob/bd4e360c73fa366f817f5aa013433e799d8cd659/share/rocm/cmake/ROCMInstallTargets.cmake#L175-L177

The symlink seems to be wrong. 
It creates the symlink in package library with link-destination `/opt/rocm/lib/..` and overwrites the original library file. The library also creates symlink in /opt/rocm/lib/ in reversed way.

Use hipFFT as example,
`\${LINK_PATH}/$<TARGET_LINKER_FILE_NAME:${TARGET}>`  is `../../lib/libhipfft.so`
`\${LINK_DIR}/${ROCM_INSTALL_LIBDIR}/$<TARGET_LINKER_FILE_NAME:${TARGET}>`  is  `/opt/rocm/hipfft/lib/libhipfft.so`
`/opt/rocm/hipfft/lib/libhipfft.so` -> `../../lib/libhipfft.so` (i.e. `/opt/rocm/lib/libhipfft.so`)
It overwrites the original library.
and hipFFT also creates `/opt/rocm/lib/libhipfft.so` -> `/opt/rocm/hipfft/lib/libhipfft.so` to lead circular symlink

lawruble13:
This issue potentially arises when the library uses `rocm_install_symlink_subdir`, which is done in all math libraries in ROCm 5.1. That said, I am also having difficulty reproducing the issue. @yhmtsai, can you confirm that this is an issue you actually encountered (as opposed to a theoretical issue)? If so, can you provide some additional information on how to reproduce the issue (e.g. a specific library & version, the commands used, environment)?

PR #93, opened 2022-04-27
Create symlinks to specific files instead of directory

lawruble13:
CPack 3.16 recurses through symlinks when generating packages, which causes file collisions with the added symlinks for the installed `cmake` folder. This creates symlinks to the files specifically, which should resolve the issue.

### DIFF
--- a/share/rocm/cmake/ROCMInstallTargets.cmake
+++ b/share/rocm/cmake/ROCMInstallTargets.cmake
@@ -329,6 +329,7 @@ function(rocm_export_targets)
 
     if(ROCM_SYMLINK_LIBS AND NOT WIN32)
         string(TOLOWER "${PROJECT_NAME}" LINK_SUBDIR)
+
         file(GENERATE OUTPUT "${CMAKE_CURRENT_BINARY_DIR}/cmake_symlink.cmake"
             CONTENT "
             set(SRC_DIR \$ENV{DESTDIR}\${CMAKE_INSTALL_PREFIX})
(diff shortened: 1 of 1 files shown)


cgmb:
Almost there. This ends up creating,

    ./opt/rocm/rocsolver/lib/cmake/rocsolver/rocsolver-targets.cmake -> ../../../lib/cmake/rocsolver/rocsolver-targets.cmake

but, the relative path ultimately just leads back to itself.

cgmb:
This ends up addressing #86 as well, right?

@saadrahim, @TorreZuk: I don't think this issue is getting enough attention. The primary reason for this change was that the generated packages for all the math libs would be invalid if generated with anything older than CMake 3.17.2. We should probably sit down and explicitly consider if this fix should go into ROCm 5.2.

The relevent CMake change that fixes things for CMake 3.17.2 and later is https://github.com/Kitware/CMake/commit/bcc5cd44ed083dbcc71692146e9d2aa3d5ba6e59.

lawruble13:
> This ends up addressing #86 as well, right?

Yes, this should address that issue as well.

TorreZuk:
Do you need @pfultz2 to approve to merge?  Maybe the code owners should have a weekly standup to get the PRs all merged that need to be done and those older than a month could be closed if not going anywhere.   Probably should create a ticket as is required if you think this should be cherry-picked for release, I don't know who sets bar for CMake version or packaging and can thus set priority @saadrahim can you or know who would?

PR #94, opened 2022-05-06
Add CODEOWNERS file

cgmb:
Code owners are automatically requested for reviews on pull requests.
That can be nice for external contributors, as it means they don't
need to guess who should be requested for review.


lawruble13 (approved):
Looks good to me, this should help ensure PRs to rocm-cmake get the requisite attention.

TorreZuk (approved):

PR #95, opened 2022-05-06
Fix use of LIBDIR

Mystro256:
This issue was found with the help of @cgmb , it looks like a lot of people have been adding this to their cmake files:

```
# Force library install path to lib (CentOS 7 defaults to lib64)
set(CMAKE_INSTALL_LIBDIR "lib" CACHE INTERNAL "Installation directory for libraries" FORCE)
```

This is not a good idea, as non-rocm builders, such as distros, will want to override the value of CMAKE_INSTALL_LIBDIR. While the use of GNUInstallDirs is a good idea in ROCMInstallTargets, as it will set sane defaults to the CMAKE_INSTALL_* variables, it also has an unintended side effect of trying to guess what to use for libdir.

We saw this issue in other components, so @frepaul had a good suggestion to set libdir prior to including GNUInstallDirs. This allows for us to have "lib" as a predicable default, but still have the advantage of overriding directories, as intended by GNUInstallDirs.

I've already implemented this logic in ROCr and OpenCL, but they haven't been merged into github yet.

### DIFF
--- a/share/rocm/cmake/ROCMInstallTargets.cmake
+++ b/share/rocm/cmake/ROCMInstallTargets.cmake
@@ -4,11 +4,14 @@
 
 cmake_policy(SET CMP0057 NEW)
 
+# Default libdir to "lib", this skips GNUInstallDirs from trying to take a guess if it's unset:
+set(CMAKE_INSTALL_LIBDIR "lib" CACHE STRING "Library install directory")
+
 include(CMakeParseArguments)
 include(GNUInstallDirs)
 include(ROCMPackageConfigHelpers)
 
-set(ROCM_INSTALL_LIBDIR lib)
+set(ROCM_INSTALL_LIBDIR ${CMAKE_INSTALL_LIBDIR})
 if(WIN32)
     set(ROCM_USE_DEV_COMPONENT OFF CACHE BOOL "Generate a devel package?")
 else()


cgmb:
I approve in principle, but there's only one development branch for rocm-cmake. A significant change like this should wait until all the fixes for ROCm 5.2.0 are sorted out.

The rocSOLVER repo has had [this awful line](https://github.com/ROCmSoftwarePlatform/rocSOLVER/blob/c398185a0df31611abbcfab99e21411c1d16e961/CMakeLists.txt#L130-L131) literally since the project was first created:
```cmake
# Force library install path to lib (CentOS 7 defaults to lib64)
set(CMAKE_INSTALL_LIBDIR "lib" CACHE INTERNAL "Installation directory for libraries" FORCE)
```

With this change, we should be able to delete the same code in:
- [rocSOLVER](https://github.com/ROCmSoftwarePlatform/rocSOLVER/blob/c398185a0df31611abbcfab99e21411c1d16e961/CMakeLists.txt#L130-L131)
- [hipSOLVER](https://github.com/ROCmSoftwarePlatform/hipSOLVER/blob/a0c56104839de86ce3042004505a5d1d855206c5/CMakeLists.txt#L164-L165)
- [rocBLAS](https://github.com/ROCmSoftwarePlatform/rocBLAS/blob/c3b37be1238d30fb604ef5eb08a8c285c4667218/CMakeLists.txt#L106-L107)
- [hipBLAS](https://github.com/ROCmSoftwarePlatform/hipBLAS/blob/7dab8cd9fba2039b45586cd5a5fae894c18fee3f/CMakeLists.txt#L154-L155)
- [rocFFT](https://github.com/ROCmSoftwarePlatform/rocFFT/blob/725f3a37a427d2c2f5d7c41b82dbabed306747c5/CMakeLists.txt#L149-L151)
- [hipFFT](https://github.com/ROCmSoftwarePlatform/hipFFT/blob/28fe3328995c52e462cd4f70e1df14e6d6238e7a/CMakeLists.txt#L148-L149)
- [rocSPARSE](https://github.com/ROCmSoftwarePlatform/rocSPARSE/blob/d24611331aa78088df7ad999e7b7d96269d24075/CMakeLists.txt#L63-L64)
- [hipSPARSE](https://github.com/ROCmSoftwarePlatform/hipSPARSE/blob/e40680d35207a64648a2fd2a1a001eba794d66a8/CMakeLists.txt#L69-L70)
- [rocALUTION](https://github.com/ROCmSoftwarePlatform/rocALUTION/blob/ef3a1f19124acd3835be4bf0a9eb8f24f9e56287/CMakeLists.txt#L63-L64)
- [rocHPCG](https://github.com/ROCmSoftwarePlatform/rocHPCG/blob/ea5946e25ccc42bede7a89928cb57feb4f90c83c/CMakeLists.txt#L52-L53)

Mystro256:
Out of curiosity, is there a reason lib64 is avoided? I assume this is just the ROCm install requirement, right? I.e. because that's where it's documented/expected to be.

Just making sure there's no technical limitation or it's one of those assumed hard coded paths somewhere. The distros wanting to package these components will want to customise the libdir path, although it's good practice to avoid any hardcoded install paths period.

cgmb:
> Out of curiosity, is there a reason lib64 is avoided? I assume this is just the ROCm install requirement, right?

I've checked https://github.com/ROCmSoftwarePlatform/rocBLAS/pull/249 and https://github.com/ROCmSoftwarePlatform/rocFFT/pull/124. Unfortunately, there was no reason given. I'd assume it was just to have the same `/opt/rocm` layout across all distros. @pfultz2 may know more, as he was around for that change.

Mystro256:
> Unfortunately, there was no reason given. I'd assume it was just to have the same /opt/rocm layout across all distros

Yes this is what I assumed too. If so, how I implemented this is fine as-is, if not, we should be wary of users/distros/community people setting a custom LIBDIR.

E.g. Fedora/RHEL/SUSE use CMAKE_INSTALL_LIBDIR="lib64", Debian/Ubuntu use CMAKE_INSTALL_LIBDIR="lib/$(uname -m)-linux-gnu".

share/rocm/cmake/ROCMInstallTargets.cmake
```diff
@@ -4,11 +4,14 @@
...
 
+# Default libdir to "lib", this skips GNUInstallDirs from trying to take a guess if it's unset:
+set(CMAKE_INSTALL_LIBDIR "lib" CACHE STRING "Library install directory")
+
 include(CMakeParseArguments)
 include(GNUInstallDirs)
 include(ROCMPackageConfigHelpers)
 
-set(ROCM_INSTALL_LIBDIR lib)
+set(ROCM_INSTALL_LIBDIR ${CMAKE_INSTALL_LIBDIR})
```
pfultz2, line 14: Why are we using cached variable for this?

pfultz2: I think it would be better to have a variable to enable non-std layout perhaps we could have `ROCM_USE_NON_STD_LAYOUT` and devops could set that instead. I dont think `CMAKE_INSTALL_LIBDIR` was meant to be used as a cached variable, so I am not sure if it might have unintended consequences. 

I would rather have the correct layout by default and then rocm can enable the non-standard layout for its packaging. We shouldn't require users to set either `ROCM_USE_STD_LAYOUT` or `CMAKE_INSTALL_LIBDIR` to get the correct layout.

pfultz2:
>  @pfultz2 may know more, as he was around for that change.

Since we dont have support for non-cmake users(like bazel), I think they are using hard-coded paths(such as `-L/opt/rocm/lib -l<some-lib>`), which switching to lib64 would break them. Once we provide pkg-config we could probably drop this.

pfultz2:
>Yes this is what I assumed too. If so, how I implemented this is fine as-is, if not, we should be wary of users/distros/community people setting a custom LIBDIR.
> 
> E.g. Fedora/RHEL/SUSE use CMAKE_INSTALL_LIBDIR="lib64", Debian/Ubuntu use CMAKE_INSTALL_LIBDIR="lib/$(uname -m)-linux-gnu".

This is another good reason why we should just have a variable to switch layouts rather than allowing `CMAKE_INSTALL_LIBDIR` to be overidden by users since it could lead to even stranger layouts.

Mystro256:
Unfortunately, not allowing arbitrary dir values isn't very distro friendly. There's no one size fit all solution as there are at least 3 or 4 interpretations of the FHS out there for LIBDIR specifically, hence why gnuinstalldirs is designed to allow overriding.

I notice most cmake based projects deal with this by not assuming any paths and often using configure_file if the path is needed in source or a script.

Also, we don't strictly need to use gnuinstalldirs, as some projects use {LIB,BIN,INCLUDE}_INSTALL_DIR (I think). I just noticed that gnuinstalldirs is already in used but the FORCE lines sort of defeat the purpose.

share/rocm/cmake/ROCMInstallTargets.cmake
Mystro256: So CMAKE_INSTALL_LIBDIR was made to be overriden, since some distros have creative libdir paths, see debian ;). The idea is that GNUInstallDirs tries to set some sensible defaults, but doesn't change the value if set. I'm not sure if the cached variable is strictly required though, I did it more just to be safe as I'm not the best with cmake.

The implementation in GNUInstallDirs is (i trimmed out some lines):

if(__system_type_for_install STREQUAL "debian")
...
  set(_LIBDIR_DEFAULT "lib/${CMAKE_LIBRARY_ARCHITECTURE}")
...
elseif(NOT DEFINED __system_type_for_install)
...
  set(_LIBDIR_DEFAULT "lib64")
...
endif()
...
if(NOT DEFINED CMAKE_INSTALL_LIBDIR)
  set(CMAKE_INSTALL_LIBDIR "${_LIBDIR_DEFAULT}" CACHE PATH "Object code libraries (${_LIBDIR_DEFAULT})")
...
endif()

cgmb: That's an interesting perspective. @Mystro256 and I were looking at this more narrowly. I certainly like the idea of the ROCm libraries following a more normal build/install process by default and just having an option to switch to the `/opt/rocm` alternative layout.

That would be a pretty significant change, though. We'd probably need to start by defaulting the option to the `/opt/rocm` layout. Then we would update our build scripts to explicitly specify that option. Finally, once everyone had specified it, we could change the default.

> Why are we using cached variable for this?

CMake itself defines CMAKE_INSTALL_LIBDIR as a cache variable when you include GNUInstallDirs. https://github.com/Kitware/CMake/blob/v3.23.1/Modules/GNUInstallDirs.cmake#L297

The reasoning is basically so that distros can pass `-DCMAKE_INSTALL_LIBDIR=<whatever>` and set it to the value that their style guide requires without having to patch our build code.

> I would rather have the correct layout by default and then rocm can enable the non-standard layout for its packaging.

I agree, but that is a big change. I don't think Jeremy's suggestion conflicts with moving to the model you suggest.

Mystro256:
Sorry I think I misunderstood your comment, email doesn't show the inline comments, so I didn't know what you meant.

share/rocm/cmake/ROCMInstallTargets.cmake
Mystro256: Personally if you ask me, I'd prefer just using GNUInstallDirs as-is, but have ROCm-CI (or whoever) override the values of CMAKE_INSTALL_PREFIX=/opt/rocm and CMAKE_INSTALL_LIBDIR=lib... but packaging didn't seem to like the idea, instead Freddy proposed the solution of just using a cached value of "lib" like I did in my patch. I think they want the default install case to fit the ROCm design to avoid confusion over customers compiling from source who don't understand that LIBDIR could be lib64 or lib or lib/somevalue.

At the end of the day though, the distros don't care as long as the paths aren't hardcoded, so those "FORCE" lines are the problem I was attempting to remedy.

saadrahim (approved):

share/rocm/cmake/ROCMInstallTargets.cmake
pfultz2: Alright, cache var is fine, we should just probably remove the `ROCM_INSTALL_LIBDIR` variable since its not needed.

saadrahim:
Please resolve conflicts and the consensus of rocm-cmake maintainers is to merge as is.

saadrahim:
Please update the version number and add to the changelog comments please.

saadrahim (changes requested):
Please update the version number and add to the changelog comments please.

saadrahim (commented):
This may need a version increment.

cgmb:
> This may need a version increment.

That is true. However, the change log already has a few entries that should be moved to the next version. I'll open a separate PR to fix them all at once. That will be a good place to discuss how to properly manage bumping the version around ROCm releases.

cgmb:
After further discussion, I've bumped the version and fixed the changelog entries. rocm-cmake 0.7.3 is going to be the one that gets released in ROCm 5.2, but the PRs that were merged today will not be part of that. I've moved the changelog entries added today to rocm-cmake 0.8.0 as well.

saadrahim (approved):

cgmb (approved):

Mystro256:
Thanks, I was a bit too busy to follow up, glad it could be merged.

There was some discussion, and I think there's better approaches, but this is a good short-term solution to avoid the "hacks".

pramenku:
Are we going to have all libs for rpm based inside "lib64" now?

This change will have library files inside "lib" path for Ubuntu but for SLES,RHEL,CentOS, it will go inside "lib64"

Are we doing this change using any feature ticket targeted to any rocm release as it's major change related to library path?

Thanks.

pramenku:
Also, hope we are communicating to all MLSE Dev too.

This change has broken hipfort which always looks for /opt/rocm-xx/lib.

Not sure what we are communicating to Application owner but everyone should be informed 1st before we should have merged the change.

http://compute-artifactory.amd.com/artifactory/rocm-qa-test-logs/test-logs/ixt-hq-54_521_2022-06-01_01:23/mlpmo_file_521_48.log

cgmb:
There was no behaviour change intended. It's merely a bug showing up in integration testing. I've opened #96 to correct the problem.

Mystro256:
Yes please merge #96 to resolve this integration bug.

PR #96, opened 2022-06-01
Ensure CMAKE_INSTALL_LIBDIR is set consistently

cgmb:
There's another `include(GNUInstallDirs)` in `ROCMCreatePackage.cmake`, so we were getting different behaviour depending on the order of includes.


cgmb:
No, but it's been about a week and the change has almost managed to percolate through the math-ci. I think we can merge.

PR #103, opened 2022-07-18
Silence RPM not found warning by adding QUIET

Maetveis:
When not building RPM packages a warning about `rpm` still shows up when a development package is requested.

"CMake Warning at /opt/rocm/share/rocm/cmake/ROCMUtilities.cmake:50 (message): 
Could not determine the version of program rpmbuild."

As this is only an optional check used to add rpm specific dependency info to the package, it should be okay to silence this warning.


cgmb (approved):
Thanks!

PR #104, opened 2022-07-20
ROCmCreatePackage: Fix pth uninstall

TedThemistokleous:
Fix for python hooks. For libraries that don't install their python api by default this seems to break workflow for the uninstall/update procedure with these libraries.


cgmb (approved):
LGTM. From [Debian's Best Packaging Practices](https://www.debian.org/doc/manuals/developers-reference/best-pkging-practices.html#best-practices-for-maintainer-scripts):

> Maintainer scripts must be idempotent. That means that you need to make sure nothing bad will happen if the script is called twice where it would usually be called once.

While `rm` will return an error status if the file does not exist, `rm -f` will not. This makes the latter idempotent.

PR #105, opened 2022-07-22
Bump version and update changelog

cgmb:
The commits on master are for `0.8.0` and are slated for a different release than those on develop.


cgmb:
It seems 285e9a2481cee99bb86cfd48b28fd4c533f20417 is going to be targeted at 5.3 after all. The version bump is thus not necessary (though not harmful). Maybe I'll drop the version bump. Then we can merge this changelog update, bump the mainline commit to the tip of develop, and fast-forward develop into master.

cgmb:
@lawruble13, please take a look. Once we get the changelog update merged, we can open a request to update mainline.

lawruble13 (approved):

saadrahim (changes requested):
Can we associate the 0.8.0 release with a particular ROCm version? Similar to the libraries release notes?

saadrahim (approved):

PR #107, opened 2022-08-04
Fix own cmake files - also create symlink from INCLUDE cmake

yhmtsai:
from 5.2.0, hip and rocm will also add cmake file into `<ROCM>/<PACKAGE>/lib/cmake` (not just `<ROCM>/lib/cmake/<PACAKGE>`)
hiprand and rocrand reports `include could not find load file: ...`
The workaround is link the missing file from `<ROCM>/lib/cmake/<PACAKGE>` to `<ROCM>/<PACKAGE>/lib/cmake`  or use `-D<pacakge_lower_case>_DIR=<ROCM>/<PACKAGE>/lib/cmake` in CMake

Issue:
When cmake find the file inside `<ROCM>/<PACKAGE>/lib/cmake`, it will create issue about cmakefile non-found.
https://github.com/RadeonOpenCompute/rocm-cmake/blob/03ec8374afec9d96afb9cc817fb9258feb84ed32/share/rocm/cmake/ROCMInstallTargets.cmake#L289-L293
only adds the INCLUDE cmake into CONFIG_PACKAGE_INSTALL_DIR, which is `lib/cmake/<PACKAGE>`
and looks for the file from current cmake folder. ( `<ROCM>/<PACKAGE>/lib/cmake`)
Thus, it can not find the file because symlink is not created.

also create the symlink for INCLUDE cmakefile.
another way is to change the path to look for the INCLUDE, but I think all cmake should be under the same folder from the rocm cmakefile structure

### DIFF
--- a/share/rocm/cmake/ROCMInstallTargets.cmake
+++ b/share/rocm/cmake/ROCMInstallTargets.cmake
@@ -263,6 +263,8 @@ function(rocm_export_targets)
 
     set(CONFIG_TEMPLATE "${CMAKE_CURRENT_BINARY_DIR}/${PACKAGE_NAME_LOWER}-config.cmake.in")
 
+    set(INCLUDED_FILES "")
+
     file(
         WRITE ${CONFIG_TEMPLATE}
         "
@@ -289,7 +291,7 @@ function(rocm_export_targets)
     foreach(INCLUDE ${PARSE_INCLUDE})
         rocm_install(FILES ${INCLUDE} DESTINATION ${CONFIG_PACKAGE_INSTALL_DIR})
         get_filename_component(INCLUDE_BASE ${INCLUDE} NAME)
-        list(APPEND PARSE_INCLUDE_NAME ${INCLUDE_BASE})
+        list(APPEND INCLUDED_FILES ${INCLUDE_BASE})
         rocm_write_package_template_function(${CONFIG_TEMPLATE} include "\${CMAKE_CURRENT_LIST_DIR}/${INCLUDE_BASE}")
     endforeach()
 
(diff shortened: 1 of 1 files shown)


share/rocm/cmake/ROCMInstallTargets.cmake
```diff
@@ -289,6 +289,7 @@ function(rocm_export_targets)
     foreach(INCLUDE ${PARSE_INCLUDE})
         rocm_install(FILES ${INCLUDE} DESTINATION ${CONFIG_PACKAGE_INSTALL_DIR})
         get_filename_component(INCLUDE_BASE ${INCLUDE} NAME)
+        list(APPEND PARSE_INCLUDE_NAME ${INCLUDE_BASE})
```
lawruble13, line 292: Can you rename the `PARSE_INCLUDE_NAME` variable you're using here, possibly to something like `INCLUDED_FILES`? `PARSE_INCLUDE_NAME` suggests that `INCLUDE_NAME` is an argument to `rocm_export_targets`.

lawruble13, line 292: Whatever variable we use should be cleared at the start of the function, to avoid accidentally using values from the calling scope.

lawruble13 (changes requested):
Thank you for pointing out the issue and suggesting a fix!

If we want to go this route to fix the problem, the noted changes should be made.

I'm of the opinion that a better route may be to resolve `CMAKE_CURRENT_LIST_DIR` using `get_filename_component(... REALPATH)` in the configuration file, although I'm not certain and will do some experimenting with the alternatives.

yhmtsai:
@lawruble13 Thanks for suggestion.
Another way might be glob all *.cmake from the source dir.
I am sure whether it will lead the circular issue I posted before again.
Also, sorry for no further reply in posted issue. I missed the notification

lawruble13 (approved):
This looks good to me, having tested & considered I think that this solution is fine, especially as this segment of code is to support backwards compatibility and will be removed eventually.

cgmb (approved):
The fix is probably going to end up in ROCm 5.5, so it might only be a single release before it's made irrelevant by the removal of the compatibility layout. Nevertheless... pushing the fix forward by a few months is still nice.

PR #108, opened 2022-08-17
Set project languages to NONE

Maetveis:
Set the languages of the projects to NONE explicitly, to avoid CMake defaulting to C and C++.

From [CMake command project() documentation](https://cmake.org/cmake/help/latest/command/project.html?highlight=project):
> By default C and CXX are enabled if no language options are given

- This speeds up the configuration by a small amount because cmake does not have to search for and verify the compilers.
- It also allows configuring the project when a C orC++ compilers are not available.
- Finally when the build of rocm-cmake is embedded in an other project (as in the case of rocRAND) it avoids errors caused by cmake selecting MSVC for the C compiler but Clang for the C++ compiler on windows.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,7 +6,7 @@ cmake_minimum_required (VERSION 3.6)
 
 set(CMAKE_INSTALL_PREFIX "/opt/rocm" CACHE PATH "")
 
-project(rocm-cmake)
+project(rocm-cmake LANGUAGES NONE)
 
 install(DIRECTORY share DESTINATION .)
 install(FILES ${CMAKE_CURRENT_BINARY_DIR}/ROCMConfigVersion.cmake DESTINATION share/rocm/cmake)
--- a/test/analyze/CMakeLists.txt
+++ b/test/analyze/CMakeLists.txt
@@ -4,7 +4,7 @@
 
 
 cmake_minimum_required (VERSION 3.5)
-project(simple)
+project(simple LANGUAGES CXX)
 
 find_package(ROCM)
 
--- a/test/findpackagecheck/CMakeLists.txt
+++ b/test/findpackagecheck/CMakeLists.txt
@@ -1,6 +1,6 @@
 
 cmake_minimum_required (VERSION 3.5)
-project(findpackagecheck)
+project(findpackagecheck CXX)
 
 message(STATUS "PKG: ${PKG}")
 message(STATUS "PKG_TARGET: ${PKG_TARGET}")
@@ -17,4 +17,3 @@ configure_file(main.cpp.in main.cpp @ONLY)
 add_executable(main ${CMAKE_CURRENT_BINARY_DIR}/main.cpp)
 target_link_libraries(main ${PKG_TARGET})
 install(TARGETS main DESTINATION bin)
-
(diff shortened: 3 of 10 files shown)


Maetveis:
Hmm, there are lot of other `project` calls in the tests, probably most of them also don't compile any code. Should I change those calls as well?

cgmb:
> * Finally when the build of rocm-cmake is embedded in an other project (as in the case of rocRAND)

How is rocRAND embedding rocm-cmake? You mean like `add_subdirectory`?

Maetveis:
> > * Finally when the build of rocm-cmake is embedded in an other project (as in the case of rocRAND)
> 
> How is rocRAND embedding rocm-cmake? You mean like `add_subdirectory`?

It's downloading, configuring and installing it via `execute_process`, when it's not found normally via `find_package` as part of `Dependencies.cmake`. There is some work so that this might change to using FetchContent or similar, but currently this is how its done.

[rocRAND code in question](https://github.com/ROCmSoftwarePlatform/rocRAND/blob/12a5d8bb17ccdf1526913a4d3aeb88eedfc7205a/cmake/Dependencies.cmake#L92-L102)

lawruble13:
Long term we would actually like to remove that code from rocRAND (and the other libraries), and have it be handled in a similar fashion to any other dependency. However this change is correct regardless, so I'm in favour. 
> Hmm, there are lot of other `project` calls in the tests, probably most of them also don't compile any code. Should I change those calls as well?

Changing those calls in the tests which don't compile anything would be appreciated, thanks.

cgmb (approved):
I'd been thinking about doing this but never got around to it, so +1 from me.

> Long term we would actually like to remove that code from rocRAND (and the other libraries), and have it be handled in a similar fashion to any other dependency.

I think there may still be a place for an optional FetchContent fallback (or something simple built on it like [CPM.cmake](https://github.com/cpm-cmake/CPM.cmake)), but generally speaking I agree. Those ad-hoc `execute_process` builds work differently in every library.

> when the build of rocm-cmake is embedded in an other project (as in the case of rocRAND) it avoids errors caused by cmake selecting MSVC for the C compiler but Clang for the C++ compiler on windows.

> It's downloading, configuring and installing it via `execute_process`, when it's not found normally via `find_package` as part of `Dependencies.cmake`.

I'm fine with this change regardless, but I don't see how rocm-cmake's configuration can affect your project's configuration when you're building it using `execute_process` like that... unless you mean that the rocm-cmake configuration itself is failing?

Maetveis:
> I'm fine with this change regardless, but I don't see how rocm-cmake's configuration can affect your project's configuration when you're building it using execute_process like that... unless you mean that the rocm-cmake configuration itself is failing?

Yes that's what's happening, because cmake is passing the compilers via environment variables (`CC` or `CXX`) to the commands executed in `execute_process` (this is not documented, but easy to verify by executing `env` via `execute_process`).
The catch is that its only setting these environment variables for the languages the project enables.
So when rocRAND (which only enables C++) is configured it needs to have `CMAKE_CXX_COMPILER=amdclang` set, so that gets passed through to rocm-cmake, but cmake will search for and default to MSVC for the C compiler resulting in an error during the configuration of rocm-cmake.

cgmb (approved):

lawruble13 (approved):

PR #109, opened 2022-08-19
Use ${CMAKE_COMMAND} for running cmake from tests

Maetveis:
Noticed this while working on #108.

The `rocm-find-program-version` test was failing at my machine because the cmake version I was using to configure rocm-cmake was not the first in my PATH.
Unlikely but possible would be that no cmake is found on the path at all, which would also lead to this test failing.

Use `${CMAKE_COMMAND}` that holds the full path to cmake to fix it.

Let me know If you think I should add this to #108, I felt like it was a separate issue.

### DIFF
--- a/test/pass/rocm-find-program-version.cmake
+++ b/test/pass/rocm-find-program-version.cmake
@@ -2,7 +2,7 @@ use_rocm_cmake()
 include(ROCMUtilities)
 
 rocm_find_program_version(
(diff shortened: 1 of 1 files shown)


lawruble13 (approved):
This looks good to me, I agree that it's a different issue.

cgmb (approved):

PR #111, opened 2022-09-16
ROCmCreatePackage: Update to prerm/postun scriptlets

solaiys:
During upgrade, the .pth files are getting removed after installing the new pkg( rpm/deb). This is because the postun/prerm scriplets has no condition check for remove/uninstall Vs upgrade.

THose .pth files should be removed only during the remove/uninstall operation.

Since the same set of scriptlets are used for both RPM and DEB, two conditions are checked for remove/uninstall and files will be removed on any one condition is satisfied ( either for RPM or DEB uninstall).

Signed-off-by: Saravanan Solaiyappan <saravanan.solaiyappan@amd.com>

### DIFF
--- a/share/rocm/cmake/ROCMCreatePackage.cmake
+++ b/share/rocm/cmake/ROCMCreatePackage.cmake
@@ -145,9 +145,14 @@ function(rocm_parse_python_syspath DIR_PATH PKG_NAME)
         "
         set_libdir(){
     ")
+    # Remove the path configuration files (.pth) only during remove/uninstall
+    # NOT during upgrade operation.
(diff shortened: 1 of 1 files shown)


raramakr (commented):
For multi rocm installation , Migrpahx package need rebuild and it will remove "rm_libdir"
This will generate an error , due to empty statements as below
if [ \"$1\" = \"remove\" ] || [ $1 -eq 0 ]; then

fi
Suggestion is to use the check inside the function rm_libdir{}

solaiys:
Closing this PR and submit a new one to address the comment given by raramakr.

PR #112, opened 2022-09-19
ROCmCreatePackage: Update to prerm/postun scriptlets

solaiys:
During upgrade, the .pth files are getting removed after installing the new pkg( rpm/deb). This is because the postun/prerm scriplets has no condition check for remove/uninstall Vs upgrade.

THose .pth files should be removed only during the remove/uninstall operation.

Since the same set of scriptlets are used for both RPM and DEB, two conditions are checked for remove/uninstall and files will be removed on any one condition is satisfied ( either for RPM or DEB uninstall).

Signed-off-by: Saravanan Solaiyappan <saravanan.solaiyappan@amd.com>

### DIFF
--- a/share/rocm/cmake/ROCMCreatePackage.cmake
+++ b/share/rocm/cmake/ROCMCreatePackage.cmake
@@ -145,9 +145,14 @@ function(rocm_parse_python_syspath DIR_PATH PKG_NAME)
         "
         set_libdir(){
     ")
+    # Remove the path configuration files (.pth) only during remove/uninstall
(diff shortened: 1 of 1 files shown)


solaiys:
@raramakr, Please review

solaiys:
Hi Reviewers & Code Owners, ( @pfultz2 , @lawruble13 , @cgmb  )
Anything pending from my side to push this change for merging?
Please review and vote.

lawruble13 (approved):
If this has been validated on all OSes, it seems fine to me.

PR #113, opened 2022-09-28
Disable color for clang tidy cache file command

tafryn:
Color codes in the output of this command can cause the subsequent FILE(MD5 ...) commands to fail.

### DIFF
--- a/share/rocm/cmake/ROCMClangTidy.cmake
+++ b/share/rocm/cmake/ROCMClangTidy.cmake
@@ -197,7 +197,8 @@ function(rocm_clang_tidy_check TARGET)
                     set(CLANG_TIDY_COMMAND_LIST \"${CLANG_TIDY_COMMAND}\")
                     set(GH_ANNOTATIONS ${ROCM_ENABLE_GH_ANNOTATIONS})
                     execute_process(
-                        COMMAND ${CMAKE_COMMAND} --build ${CMAKE_CURRENT_BINARY_DIR} --target ${BASE_SOURCE}.i
+                        COMMAND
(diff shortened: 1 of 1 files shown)


share/rocm/cmake/ROCMClangTidy.cmake
```diff
@@ -197,7 +197,8 @@ function(rocm_clang_tidy_check TARGET)
                     set(CLANG_TIDY_COMMAND_LIST \"${CLANG_TIDY_COMMAND}\")
                     set(GH_ANNOTATIONS ${ROCM_ENABLE_GH_ANNOTATIONS})
                     execute_process(
-                        COMMAND ${CMAKE_COMMAND} --build ${CMAKE_CURRENT_BINARY_DIR} --target ${BASE_SOURCE}.i
+                        COMMAND
+                            ${CMAKE_COMMAND} --build ${CMAKE_CURRENT_BINARY_DIR} --target ${BASE_SOURCE}.i -e COLOR=OFF
```
pfultz2, line 201: Where did you find documentation on the `-e COLOR=OFF` flag? And does this work with cmake 3.16? I'll try to run this locally.

Unfortunately, I dont think we have tests for tidy.

pfultz2 (approved):
I checked locally and this works.

share/rocm/cmake/ROCMClangTidy.cmake
tafryn: The trailing options are forwarded to make where `-e` is used for environment overrides.

cgmb:
This appears to be the issue in https://www.reddit.com/r/ROCm/comments/z3t46p/issues_with_building_rocm_from_aur_arch/

Merging based on @pfultz2's review.

PR #118, opened 2022-11-23
Add versioning for DLLs on Windows

lawruble13:


share/rocm/cmake/ROCMSetupVersion.cmake
```diff
@@ -156,7 +156,7 @@ function(rocm_get_patch_version OUTPUT)
 endfunction()
 
 function(rocm_set_soversion LIBRARY_TARGET SOVERSION)
-    if(NOT WIN32 AND NOT APPLE)
+    if(NOT APPLE)
```
cgmb, line 159: Why do we bother with `if(NOT APPLE)`?

test/pass/rocm-list-split.cmake
```diff
@@ -1,5 +1,5 @@
 use_rocm_cmake()
-include(ROCMInstallTargets)
+include(ROCMUtilities)
```
cgmb, line 2: Isn't this an indication that moving that function is a potentially breaking change for rocm-cmake users?

share/rocm/cmake/ROCMVersionDLL.cmake
```diff
@@ -0,0 +1,58 @@
...
+set(_ROCMVersionDLL_LIST_DIR "${CMAKE_CURRENT_LIST_DIR}")
+
+define_property(
+    TARGET PROPERTY DLL_FILE_VERSION
+    BRIEF_DOCS "This target property records the version of the generated DLL file."
+    FULL_DOCS "The value stored in this property will be set as the FileVersion property of the DLL."
+)
+
+set(ROCM_DLL_VERSION "$ENV{ROCM_DLL_VERSION}" CACHE STRING "The version to add to DLLs by default.")
+set(ROCM_DLL_FILE_VERSION "$ENV{ROCM_DLL_FILE_VERSION}" CACHE STRING "The version of the file to attach to DLLs.")
```
cgmb, line 14: I don't think this makes sense as a global option. It's really a property of the library itself and I don't think we want to bake in the assumption that there's only one library built per CMake invokation. There's a lot of ways that assumption could be violated, e.g., building libraries via `add_subdirectory`.

share/rocm/cmake/ROCMVersionDLL.cmake
```diff
@@ -0,0 +1,58 @@
...
+            set(DLL_PRODUCT_VERSION "${${PROJECT_NAME}_VERSION}")
+        elseif(ROCM_DLL_PRODUCT_VERSION STREQUAL "")
+            set(DLL_PRODUCT_VERSION "${ROCM_DLL_VERSION}")
+        else()
+            set(DLL_PRODUCT_VERSION "${ROCM_DLL_PRODUCT_VERSION}")
+        endif()
+        if(
+            NOT DEFINED DLL_FILE_VERSION OR NOT DEFINED DLL_PRODUCT_VERSION OR
+            DLL_FILE_VERSION STREQUAL "" OR DLL_PRODUCT_VERSION STREQUAL ""
+        )
```
cgmb, line 45: Is this essential complexity? That is, do we actually need to support multiple different ways of setting these properties? Sometimes it's better to leave out 'convenience' features. The more ways something can be done, the more possibilities that you need to understand when debugging when something goes wrong...

cgmb (commented):
Interesting. I'm not familiar with this DLL metadata. Is this orthogonal to the XML manifest? And is there a ticket associated with this work?

eidenyoshida:
@lawruble13 wondering if there is still an ask for this functionality?

eidenyoshida:
@apwojcik do you have any comments or feedback on this approach to adding versioning for DLLs in Windows?

eidenyoshida:
Closing stale ticket

PR #123, opened 2023-02-02
ASAN Packaging

lawruble13:


lawruble13:
@cgmb @pfultz2 Can we get this looked at as well, please?

cgmb (commented):
I disagree in principle with defining the filesystem layout in rocm-cmake, but we can make it work so I can live with it. Just had to say it once so it was on record. I'm not going to die on this hill.

For a more substantive response: isn't `rocfft_rtc_helper` going to create a conflict between the normal package and the asan package since both will be installed to `$ROCM_PATH/bin`? There should probably be handling for libexec binaries and rocfft should probably use it.

pfultz2 (approved):

cgmb (approved):
We can deal with any conflicts between rocfft and rocfft-asan in a followup. LGTM.

PR #125, opened 2023-03-14
Add function to get local build targets

lawruble13:


eidenyoshida:
Should we investigate addressing this at a different level such as hipconfig?

lawruble13:
Yes, having this in the ROCmCMakeBuildTools would always be more of a workaround than an actual solution to the issue. I had left this PR up for easy reference as a possible CMake implementation while other solutions were being discussed.

cgmb:
Let's close this out, then. The PR will always be available as a reference.

Issue #49, opened 2021-06-17
find_package (ROCM is confusing


saadrahim:
Invoking 
find_package( ROCM 0.5 CONFIG QUIET PATHS ${ROCM_PATH} /opt/rocm )
to include rocm-cmake is confusing. 

We should consider changing the name to ROCM-CMAKE.

lawruble13:
@pfultz2 Can you weigh in on this? Using the name `ROCM` is confusing, especially with the version numbers which don't match the ROCm version number, and we have the opportunity to make a breaking change with ROCm 5.0.

nartmada:
Hi @saadrahim, is there any further work required for this ticket?  Can I close it?  Thanks.

cgmb:
It is `find_package(ROCmCMakeBuildTools)` in ROCm 6.1 and later, although `find_package(ROCM)` remains available for backwards-compatibility.

PR #128, opened 2023-05-11
Rename CMake package to ROCmCMakeBuildTools

lawruble13:
Having the name of this CMake package be `ROCM` is consistently confusing to users who expect `find_package(ROCM [...])` to add ROCm packages to their build, instead of tools that we use to build ROCm (see #49). Renaming the CMake project to ROCmCMakeBuildTools makes the purpose of this package clearer.


saadrahim (approved):

share/rocm/cmake/ROCMConfig.cmake
```diff
@@ -1,6 +1,9 @@
 # ######################################################################################################################
-# Copyright (C) 2017 Advanced Micro Devices, Inc.
+# Copyright (C) 2023 Advanced Micro Devices, Inc.
 # ######################################################################################################################
 
-list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_LIST_DIR})
-include(ROCMChecks)
+message(INFO
```
cgmb, line 5: This should be of type [`DEPRECATION`](https://cmake.org/cmake/help/v3.0/command/message.html#general-messages).

share/rocm/cmake/ROCMConfig.cmake
```diff
@@ -1,6 +1,9 @@
...
+    "Use of find_package(ROCM) is deprecated as of <version>, and will be removed in <version>, "
+    "please switch to find_package(ROCmCMakeBuildTools)."
+)
+find_package(ROCmCMakeBuildTools PATHS .)
```
cgmb, line 9: My understanding is that the `PATHS` property is for searching static paths. If want to search a path that is relative to some other package, the appropriate property is `HINTS`. I'd also be more comfortable searching `"${CMAKE_CURRENT_LIST_DIR}"` rather than `.`.

share/rocm/cmake/ROCMConfig.cmake
```diff
@@ -1,6 +1,9 @@
 # ######################################################################################################################
-# Copyright (C) 2017 Advanced Micro Devices, Inc.
+# Copyright (C) 2023 Advanced Micro Devices, Inc.
 # ######################################################################################################################
 
-list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_LIST_DIR})
-include(ROCMChecks)
+message(INFO
+    "Use of find_package(ROCM) is deprecated as of <version>, and will be removed in <version>, "
+    "please switch to find_package(ROCmCMakeBuildTools)."
```
cgmb, line 7: This should be edited for brevity. It's sufficient to say that it is deprecated and what to use instead.

cgmb (commented):
Why did the build fail before the tests were changed? Our tests passing without requiring any changes would be proof that this is being done in a backwards-compatible manner. We don't have to release (or even merge) like that, but it would be nice to see it pass once without test changes, if possible.

lawruble13:
The tests failed prior to switching the package name due to the [`-Werror=dev` argument](https://github.com/lawruble13/rocm-cmake/blob/4b211a55bd5c7b71e49aa61940fbb943b0709e60/test/test.cmake#L88) passed to CMake when running the tests. I agree though that there should be tests validating that the backwards compatibility measures do actually work.

share/rocm/cmake/ROCMConfig.cmake
lawruble13: Thanks for the correction, updated.

share/rocm/cmake/ROCMConfig.cmake
lawruble13: Agreed on both counts, updated.

keryell (commented):
I can see in the description of this PR: "No description provided." which is not very inclusive for an open-source project. :-)
Any rationale for this PR?
At least having the information from the `CHANGELOG.md` could be a good start.
Feel free to put some comments in the CMake files themselves too to help future readers/users.

saadrahim:
> I can see in the description of this PR: "No description provided." which is not very inclusive for an open-source project. :-) Any rationale for this PR? At least having the information from the `CHANGELOG.md` could be a good start. Feel free to put some comments in the CMake files themselves too to help future readers/users.

@lawruble13 can you address this request?

keryell (approved):
Thanks for the clarification.

lawruble13:
I changed the `DEPRECATION` to a `NOTICE` so that we can start letting people know about this before we start to throw warnings/errors at them depending on their settings.

share/rocm/cmake/ROCMConfig.cmake
```diff
@@ -2,7 +2,7 @@
 # Copyright (C) 2023 Advanced Micro Devices, Inc.
 # ######################################################################################################################
 
-message(DEPRECATION
+message(NOTICE
     "Use of find_package(ROCM) is deprecated, please switch to find_package(ROCmCMakeBuildTools)."
```
keryell, line 6: Then, you probably need to change the message to something like
```
    "Use of find_package(ROCM) will be deprecated soon, please switch to find_package(ROCmCMakeBuildTools)."
```

cgmb (commented):
> I changed the `DEPRECATION` to a `NOTICE` so that we can start letting people know about this before we start to throw warnings/errors at them depending on their settings.

The `NOTICE` category wasn't added until CMake 3.15. You could use `STATUS`, but I would suggest leaving it as a `DEPRECATION`. Users can set [`CMAKE_ERROR_DEPRECATED`](https://cmake.org/cmake/help/v3.27/variable/CMAKE_ERROR_DEPRECATED.html#variable:CMAKE_ERROR_DEPRECATED) or [`CMAKE_WARN_DEPRECATED`](https://cmake.org/cmake/help/v3.27/variable/CMAKE_WARN_DEPRECATED.html#variable:CMAKE_WARN_DEPRECATED) to choose how they want deprecations to be treated. If you don't want to deprecate it yet, another option would be to omit the message entirely.

The updated deprecation message is better than before, but I think you would benefit from reviewing deprecation messages from project like rust, libstdc++ or Qt so that you can match their style and tone.

PR #129, opened 2023-05-23
Fix issues when using rocm test module from a subdirectory

pfultz2:
The `rocm_defer` workaround actually called the function for each subdirectory which causes an error in cmake when generating the same file multiple times. Also, updated the tests to use a subdirectory to more closely match how this would be implemented.


lawruble13 (approved):
This seems reasonable to me.

PR #132, opened 2023-06-12
SWDEV-405323 - Add suffix asan to cpack package name for ASAN builds

raramakr:
In SLES, the asan package name doesn't have the suffix asan. Append cpack package name with suffix asan


nunnikri (approved):

cgmb (approved):
LGTM, but @lawruble13 knows this code better than me.

lawruble13 (approved):
This change looks okay, although I confess to being somewhat confused as to why the original code produced incorrect behaviour. I will be investigating that further, but this change is OK as far as I'm concerned.

lawruble13:
Didn't take that much investigation, the  [`CPACK_RPM_MAIN_COMPONENT`](https://cmake.org/cmake/help/latest/cpack_gen/rpm.html#variable:CPACK_RPM_MAIN_COMPONENT) variable creates exactly this behaviour.

PR #135, opened 2023-06-28
make Clang-Tidy and CppCheck optional

apwojcik:
With this PR, CMake is only generating additional clang-tidy and cppcheck targets if the tools are installed in the system (can be found).  As an option, the PR provides the method to switch the checks off even if the tools are present.

On Windows, the Clang-Tidy and CppCheck are not installed by default. If missing, the tools block successful build configuration. On the other hand, when one of them is available, the CMake generates a vast number of targets related to code checks. Those targets pollute the generated project and are exposed to an IDE project view, and the "real" targets are hard to find.


pfultz2 (approved):
LGTM, just need to fix the linter issue.

PR #136, opened 2023-07-10
SWDEV-410105 - For ASAN packages add dependency to rocm-core-asan rather than rocm-core

raramakr:


share/rocm/cmake/ROCMCreatePackage.cmake
```diff
@@ -131,8 +131,14 @@ macro(rocm_package_add_rocm_core_dependency)
...
     # to be removed in one step by removing rocm-core
     if(ROCM_DEP_ROCMCORE)
-        rocm_join_if_set(", " CPACK_DEBIAN_PACKAGE_DEPENDS "rocm-core")
-        rocm_join_if_set(", " CPACK_RPM_PACKAGE_REQUIRES "rocm-core")
+        if(ENABLE_ASAN_PACKAGING)
+            # For ASAN packages, add dependency to rocm-core-asan
+            set(ROCM_CORE_PKG "rocm-core-asan")
+        else()
+            set(ROCM_CORE_PKG "rocm-core")
+        endif()
```
cgmb, line 139: Given that this is a macro (not a function), do we have to be careful about defining new variables? This variable might be better named `_rocm_core_pkg` and then unset after it is no longer needed.

raramakr: Thank you @cgmb  for the review comments. Addressed the same in the latest patch

cgmb (approved):

lawruble13 (approved):

PR #138, opened 2023-07-13
Add options to control install behaviour in ASAN builds

lawruble13:


share/rocm/cmake/ROCMInstallTargets.cmake
```diff
@@ -119,12 +129,16 @@ endfunction()
...
 
 function(rocm_install_targets)
-    set(options)
+    set(options PRIVATE)
     set(oneValueArgs PREFIX EXPORT COMPONENT)
     set(multiValueArgs TARGETS INCLUDE)
 
     cmake_parse_arguments(PARSE "${options}" "${oneValueArgs}" "${multiValueArgs}" ${ARGN})
 
+    if(PARSE_PRIVATE AND ENABLE_ASAN_PACKAGING)
```
lawruble13, line 138: I want to note as I put this forward that I'm not particularly happy with this implementation, and would greatly appreciate feedback on it.

pfultz2:
So `PRIVATE` means its not a user facing file, and is only used internally by the library. As such, it should be installed when using asan and not asan builds. For `PRIVATE` installs it should go into `lib/<project-name>/` or `lib/asan/<project-name>`. This way the relative path to the file should be the same regardless if its asan or non-asan.

lawruble13:
> So `PRIVATE` means its not a user facing file, and is only used internally by the library. As such, it should be installed when using asan and not asan builds. 

Ah, I misunderstood what you had said in our last discussion. Should there then be a flag to force a target to be ignored in the ASAN build, in addition to this `PRIVATE` flag?

share/rocm/cmake/ROCMInstallTargets.cmake
```diff
@@ -146,6 +156,11 @@ function(rocm_install_targets)
         set(INCLUDE_INSTALL_DIR ${CMAKE_INSTALL_INCLUDEDIR})
     endif()
 
+    if (PARSE_PRIVATE)
+        set(BIN_INSTALL_DIR ${BIN_INSTALL_DIR}/${PROJECT_NAME})
```
pfultz2, line 160: This should be `${LIB_INSTALL_DIR}/${PROJECT_NAME}`.

share/rocm/cmake/ROCMInstallTargets.cmake
```diff
@@ -146,6 +156,11 @@ function(rocm_install_targets)
...
+        set(LIB_INSTALL_DIR ${LIB_INSTALL_DIR}/${PROJECT_NAME})
```
pfultz2, line 161: I dont know if it makes sense to have a private lib directory. Either way, I think this would point to the same directory.

pfultz2:
> Should there then be a flag to force a target to be ignored in the ASAN build, in addition to this PRIVATE flag?

No, we should just ignore everything thats not a shared object by default in ASAN builds. Then the `PRIVATE` flag can install the file into `lib/${PROJECT_NAME}` or `lib/asan/${PROJECT_NAME}` for runtime components when ASAN is enabled.

PR #140, opened 2023-08-04
Fix extra make flags passed for clang tidy

pfultz2:
In newer cmake it doesn't implicitly pass the flags along(this seemed to be an undocumented feature). Either way, I added `--` to fix this. 

Also, updated the tests to find these issues in the future.


pfultz2:
@lawruble13 @cgmb Please review. For some reason I can't assign you as reviewers. Thanks!

lawruble13 (approved):
LGTM

PR #143, opened 2023-09-01
More fixes to ROCMTest module

pfultz2:
This fixes problems with how the component was created and installed. Also I added more tests to catch these issues. The tests will now check that the test package contains test files.


pfultz2:
@cgmb @lawruble13 Any comments?

cgmb (approved):
I'm not deeply familiar with the design of the ROCMTest module, but these changes look reasonable to me. @lawruble13 more thoroughly reviewed the original module and may have more insights, but LGTM.

PR #144, opened 2023-09-06
Dont require sphinx to be installed

pfultz2:
This should only be required when building the documentation. There is no reason to require it when users just wants to build the component.

### DIFF
--- a/share/rocm/cmake/ROCMSphinxDoc.cmake
+++ b/share/rocm/cmake/ROCMSphinxDoc.cmake
@@ -12,8 +12,7 @@ find_program(
     NAMES sphinx-build
     HINTS "$ENV{SPHINX_DIR}"
     PATH_SUFFIXES bin
-    DOC "Sphinx documentation generator"
-    REQUIRED)
+    DOC "Sphinx documentation generator")
 
 find_program(
     DOXYGEN_EXECUTABLE
--- a/share/rocm/cmake/ROCMSphinxDoc.cmake
+++ b/share/rocm/cmake/ROCMSphinxDoc.cmake
@@ -60,10 +60,10 @@ function(rocm_add_sphinx_doc SRC_DIR)
         list(APPEND VARS -A "${VAR}")
     endforeach()
 
-    if(PARSE_USES_DOXYGEN)
-        set(USES_DOXYGEN -D "doxygen_executable=${DOXYGEN_EXECUTABLE}")
+    if(PARSE_USE_DOXYGEN)
+        set(USE_DOXYGEN -D "doxygen_executable=${DOXYGEN_EXECUTABLE}")
     else()
-        set(USES_DOXYGEN)
+        set(USE_DOXYGEN)
     endif()
 
     if(NOT TARGET sphinx-${BUILDER})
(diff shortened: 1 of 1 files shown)


cgmb (commented):
Good point. However, there should be a check added to `rocm_add_sphinx_doc` to ensure it was found before using the variable.

pfultz2:
> here should be a check added to rocm_add_sphinx_doc to ensure it was found before using the variable.

CMake should already error when the variable is set to not found. Either way, how would we write such an error?

cgmb:
It's not directly related to this change, but I noticed a typo in the existing code where the option is named `USE_DOXYGEN` but the later code expects `USES_DOXYGEN`.

cgmb:
> CMake should already error when the variable is set to not found.

CMake won't emit an error. The error will occur at build time rather than at configuration time. Make will attempt to execute a program named `SPHINX_EXECUTABLE-NOTFOUND`, which will typically result in the error `SPHINX_EXECUTABLE-NOTFOUND: Command not found`.

I agree with you that including ROCMSphinxDoc shouldn't automatically cause sphinx to be required. However, allowing configuration to complete with a nonsense command for one of the targets doesn't seem right. I see two options:

1. `rocm_add_sphinx_doc` could emit an error `if(PARSE_USE_DOXYGEN AND NOT DOXYGEN_EXECUTABLE)` or `if(NOT SPHINX_EXECUTABLE)`. Users would have to guard calls to `rocm_add_sphinx_doc` with `if(SPHINX_EXECUTABLE AND DOXYGEN_EXECUTABLE)` if they set `USE_DOXYGEN` or just `if(SPHINX_EXECUTABLE)` if they do not use doxygen.
2. `rocm_add_sphinx_doc` could return without doing anything `if(PARSE_USE_DOXYGEN AND NOT DOXYGEN_EXECUTABLE)` or `if(NOT SPHINX_EXECUTABLE)`. Users would not need to guard calls to `rocm_add_sphinx_doc`, but the doc target might not exist.

pfultz2:
> CMake won't emit an error. The error will occur at build time rather than at configuration time

We want the error to happen at build time not at configuration time. 

> Users would have to guard 

Yea we dont want to require that. Unless a component has a CI that builds with a minimal dependencies(we dont have that), there nothing to enforce the users are adding the guard.

> Users would not need to guard calls to rocm_add_sphinx_doc, but the sphinx-${BUILDER} and ${PROJECT_NAME}-sphinx-${BUILDER} targets might not be created if the required tools are not available.

Yea and I think `SPHINX_EXECUTABLE-NOTFOUND: Command not found` is much clearer to what when wrong then having it say `*** No rule to make target 'sphinx-html'.`. 

Also there would only be an error if calling `make sphinx-html`. Calling `make doc` will just silently not do anything, which is even worse.

pfultz2:
Another idea, we could add the `sphinx-${BUILDER}` targets as `add_custom_target(sphinx-${BUILDER} ${CMAKE_COMMAND} -E echo "Sphinx executable not found" COMMAND ${CMAKE_COMMAND} -E false)`.

cgmb:
In the interest of time, maybe let's proceed with the parts that everyone is agreed on and defer error-handling improvements for later. If you roll the PR back to just 7d069fb2985735d671c40274be08a293d5df45cd and 99ab7b4b959b81f3cd78c5725a9b44ee764ead46, I'd approve. Those changes are necessary but the rest are merely nice to have.

pfultz2:
Done!

cgmb (approved):

pramenku:
@lawruble13 @pfultz2 

We need PR merge in https://github.com/RadeonOpenCompute/rocm-cmake/tree/release-staging/rocm-rel-6.0 branch too so that it will come to 6.0 Mainline.
please help.

PR #147, opened 2023-09-12
Improve error reporting for missing sphinx/doxygen executable

pfultz2:


cgmb (approved):
I still think it makes the most sense to emit the error at configuration time, but I'm not going to die on that hill when you've implemented a reasonable alternative.

PR #148, opened 2023-09-20
Allow separate source and config dirs in sphinx

MathiasMagnus:
rocm-docs-core up until now assumed that during Sphinx documentation build the configure and source directories were the same. To allow use cases where project maintainers don't want to litter the source tree with build side-effects (understandable), this assumption is being cleaned up within rocm-docs-code (see [here](https://github.com/RadeonOpenCompute/rocm-docs-core/pull/409)).

This is the rocm-cmake side of supporting such use cases, allowing setting the `-c` flag of Sphinx. `conf.py` is searched either within this folder, or if not set, within the source tree, as was the case until now.

### DIFF
--- a/share/rocm/cmake/ROCMSphinxDoc.cmake
+++ b/share/rocm/cmake/ROCMSphinxDoc.cmake
@@ -27,13 +27,13 @@ mark_as_advanced(DOXYGEN_EXECUTABLE)
 
 function(rocm_add_sphinx_doc SRC_DIR)
     set(options USE_DOXYGEN)
-    set(oneValueArgs BUILDER OUTPUT_DIR)
+    set(oneValueArgs BUILDER CONFIG_DIR OUTPUT_DIR)
     set(multiValueArgs DEPENDS VARS TEMPLATE_VARS)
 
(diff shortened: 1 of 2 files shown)


lawruble13 (approved):
This looks good to me, @cgmb @pfultz2 any other feedback?

share/rocm/cmake/ROCMSphinxDoc.cmake
```diff
@@ -79,6 +89,7 @@ function(rocm_add_sphinx_doc SRC_DIR)
         ${PROJECT_NAME}-sphinx-${BUILDER}
         COMMAND
             "${SPHINX_EXECUTABLE}"
+            ${CONFIG_DIR_ARG}
```
cgmb, line 92: Should this be quoted?

cgmb (approved):
Seems reasonable to me.

share/rocm/cmake/ROCMSphinxDoc.cmake
MathiasMagnus: While it's going to need extra testing, for the time being, it worked for me on Linux when the source tree was accessed through a symlink with a space in it.

PR #150, opened 2023-10-11
Cherrypick Sphinx fixes to ROCm 6.0

lawruble13:
* Dont require sphinx to be installed

* Dont require doxygen

* Fix doxygen variable name


cgmb:
I think we can just push 5a34e72d9f113eb5d028e740c2def1f944619595 to release-staging/rocm-rel-6.0. That just has the changelog update and another bugfix in addition to this.

pfultz2 (approved):

PR #152, opened 2023-10-11
Cherrypick to 6.0: Rename CMake package to ROCmCMakeBuildTools 

lawruble13:
For discussion, see: Rename CMake package to ROCmCMakeBuildTools (#128).

Having the name of this CMake package be ROCM is consistently confusing to users who expect find_package(ROCM [...]) to add ROCm packages to their build, instead of tools that we use to build ROCm (see #49). Renaming the CMake project to ROCmCMakeBuildTools makes the purpose of this package clearer.

In order to have as smooth a transition as possible, starting this as a notice, will make into a deprecation/warning in ~6.2, error sometime later, remove in 7.0.


cgmb (approved):
Thanks for handling this.

cgmb:
Oh. I did want to note that making a cherry-pick, you can use the `-x` option to have git add a message noting the original commit it was picked from. Not a big deal, but I think it's nice information to include. No need to change this one, though. Maybe give it a try next time.

lawruble13:
> Oh. I did want to note that making a cherry-pick, you can use the `-x` option to have git add a message noting the original commit it was picked from.

Thanks, I didn't know that! I'll be sure to use it in the future.

cgmb:
@lawruble13, I reverted the cherry-pick due to failures in other components. This feature needs more time to bake. It's too late in the ROCm 6.0 development cycle to make changes that are potentially breaking for other components.

There wasn't a fundamental need to ship this in ROCm 6.0, aside from the paperwork involved with deferring it. If someone gives you a hard time on the deferral, refer them to me.

PR #155, opened 2023-11-09
Correct path for ROCMConfig redirect

lawruble13:
The previous redirect in ROCMConfig.cmake did not correctly specify the path to ROCmCMakeBuildTools. This caused failures if the user only specified the path to ROCM, (via something like `ROCM_DIR`), as there were then no correct paths to search for ROCmCMakeBuildTools. I want to create a test for this, but I'm not quite sure how yet, it's in progress.


share/rocm/cmake/ROCMConfig.cmake
```diff
@@ -2,4 +2,11 @@
...
 
-find_package(ROCmCMakeBuildTools HINTS "${CMAKE_CURRENT_LIST_DIR}")
+get_filename_component(_new_rocmcmakebuildtools_path "${CMAKE_CURRENT_LIST_DIR}" DIRECTORY)
+get_filename_component(_new_rocmcmakebuildtools_path "${_new_rocmcmakebuildtools_path}" DIRECTORY)
+# two directories up is sufficient for windows search, but linux search requires the share directory
+get_filename_component(_new_rocmcmakebuildtools_path_linux "${_new_rocmcmakebuildtools_path}" DIRECTORY)
+
+find_package(ROCmCMakeBuildTools REQUIRED HINTS
+    "${_new_rocmcmakebuildtools_path}"
+    "${_new_rocmcmakebuildtools_path_linux}")
```
cgmb: Should this be `find_package(ROCmCMakeBuildTools HINTS`?

lawruble13: @cgmb I've updated it to pass through the `REQUIRED` and `QUIET` arguments if provided, and the version argument if provided as well. Can you take another look?

lawruble13: As an aside, the `REQUIRED` flag was used originally as `ROCmCMakeBuildTools` should always be installed alongside this file, and so the `REQUIRED` was effectively a short-cut for erroring out if it couldn't be found (indicating that something is wrong).

lawruble13: I looked at your comment prior to your edit and saw that you probably intended to suggest `find_dependency`. I have updated to use this, as it obviates the previous work I did to forward the arguments.

lawruble13:
Note that the failures on macOS are likely unrelated, as the develop branch is also failing on macOS, which should be investigated.

share/rocm/cmake/ROCMConfig.cmake
cgmb: Good for catching that. Sorry for being unclear. I tried to add the HINTS with an edit and it seems I accidentally overwrote the core point via a copy/paste error.

cgmb (approved):
You should perhaps unset your temporary variables at the end of the config script, but this looks good to me.

PR #159, opened 2023-11-21
[ROCM 5.7] fix default installation path on Windows

apwojcik:
The PR is for the 5.7 branch because we need it for the ROCm 5.7 release on Windows.

It is unusual on Windows to install to `C:\Program Files` manually. It requires Administrator privilege, which is not standard or not easily granted. Only applications managed by the Windows installer are allowed to install in that directory, and `cmake install` is not one of them.
Installing tests in a custom directory given to CMake is also impossible when `HEADER_INSTALL_PREFIX` is set to `C:/Program Files`. Moreover, the directory `C:\Program Files` is for 64-bit applications only. In the case of 32-bit, the installer should use the `C:\Program Files (x86)` directory instead.

The separate PRs will add that change to the current development branch.


apwojcik:

PR #160, opened 2023-11-21
[ROCM 5.7] Make installing tests optional for ROCm component

apwojcik:
The PR is for the 5.7 branch because we need it for the ROCm 5.7 release on Windows.

It is unusual for a component to install a test suite when only deployment binaries are required. The PR makes the installation of tests optional for such an ROCm component. Also, the PR adds installation of PDB files (debugging database) for tests on Windows (automatically for Debug build types only).

The separate PRs will add that change to the current development branch.


pfultz2:
Installing tests are already optional. When doing `cmake --build . --target install`, it wont install the tests.

apwojcik:

PR #161, opened 2023-11-21
fix default installation path on Windows

apwojcik:
It is unusual on Windows to manually install in the `C:\Program Files` directory. It requires Administrator privilege, which is not standard or not easily granted. Only applications managed by the Windows installer are allowed to install in that directory, and `cmake install` is not one of them.
Installing tests in a custom directory provided as a parameter to CMake is also impossible when `HEADER_INSTALL_PREFIX` is set to `C:/Program Files`. Moreover, the directory `C:\Program Files` is for 64-bit applications only. In the case of 32-bit, the installer should use the `C:\Program Files (x86)` directory instead.


apwojcik:

PR #165, opened 2024-01-05
Add rpath for installed tests

pfultz2:


cgmb (changes requested):
The addition of an rpath/runpath should always be optional, since the use of those properties is discouraged by the packaging guidelines on distros like Debian.

cgmb (commented):
My mistake. It is still optional with `CMAKE_SKIP_RPATH`.

cgmb (approved):

PR #169, opened 2024-02-08
Update version number and changelog for ROCm 6.1

cgmb:


lawruble13 (approved):
Should we explicitly note that `find_package(ROCM [...])` will still work via a compatibility layer?

Aside from that question, this looks good to me, so unless you feel that needs adding I'm okay with this being merged.

PR #175, opened 2024-03-05
Rename check target to rocm-cmake-check

powderluv:
This is so that we can add_subdirectory() this project and not clash with LLVM's check target.

TEST: running before and after has same number of check failures (3) on my system


pfultz2 (changes requested):
This should be named `check` so its consistent across projects.

cgmb:
How about rocm-cmake only creates the check target if `PROJECT_IS_TOP_LEVEL` is true?

PR #178, opened 2024-03-13
Install executables to bin/asan for ASAN builds

lawruble13:
There is a request to install executables for ASAN builds to a separate location (specifically this is in the context of test executables, but we don't have the ability to filter that).
We've had some discussions on library-private executables, and I'd appreciate if anyone has a more robust idea than this, but here's an initial extremely simple take.


pfultz2 (approved):
I think this looks fine for now. I dont see this as a problem for the private executables currently installed in `bin` since they were already a different path so it was never used for asan builds.

Later on, we can try and add the `PRIVATE` flag so the private executables dont get installed into `bin` or `bin/asan`, and also get adjusted for asan builds.

share/rocmcmakebuildtools/cmake/ROCMInstallTargets.cmake
```diff
@@ -137,7 +137,7 @@ function(rocm_install_targets)
         set(LIB_INSTALL_DIR ${PARSE_PREFIX}/${ROCM_INSTALL_LIBDIR})
         set(INCLUDE_INSTALL_DIR ${PARSE_PREFIX}/${CMAKE_INSTALL_INCLUDEDIR})
     elseif(ENABLE_ASAN_PACKAGING)
-        set(BIN_INSTALL_DIR ${CMAKE_INSTALL_BINDIR})
+        set(BIN_INSTALL_DIR ${CMAKE_INSTALL_BINDIR}/asan)
```
nunnikri, line 140: Is this a new requirement came from llvm/asan team?

PR #182, opened 2024-03-28
Packaging for static components

lawruble13:
The static packages were previously named inconsistently, with clients packages getting `-static` and the develop package remaining unchanged. Additionally, the runtime package is still generated for static builds, and the license file is installed there. This changes that behaviour to be more correct.


CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Change Log for rocm-cmake
 
+## [(Unreleased) rocm-cmake 0.13.0 for ROCm 6.2.0]
+### Changed
+- ROCmCreatePackage: Accepts a suffix parameter, automatically generating it for static or ASAN builds.
+  - Package names will no longer be pulled from `CPACK_<GEN>_PACKAGE_NAME`
+  - Runtime packages will no longer be generated for static builds
```
cgmb, line 7: Will this cause problems for rocBLAS? The library cannot be fully static yet...

lawruble13: It's my understanding that the rocBLAS build on our testing is successful with this change.

cgmb: Just to summarize our offline discussion: even if rocBLAS needs extra files (for now), it's not the responsibility of the package system to provide them. The point of the static libraries is to not need to use the packaging system at run-time.

cgmb (approved):

PR #187, opened 2024-05-03
reorganize doc folder content for standard build

randyh62:
Moved content up from ./docs/src folder into ./docs folder
created ./docs/sphinx folder
corrected paths in CMakeLists.txt and other files as needed

This is not a reorg for diataxis


pfultz2:
I would like the source files in a seperate directory so it doesnt interfere with other build files that are needed.

PR #202, opened 2024-06-18
HOTFIX: Fix names for header-only components when built statically

lawruble13:
Statically built packages may have different dependencies than the non-static packages, even for header only components whose contents are otherwise identical. As such, the generated packages should have separate names.


lawruble13:
Duplicate of #204 targeting ROCm 6.2

PR #203, opened 2024-06-18
change ordering of suffixes

lawruble13:
The ordering of suffixes is inconsistent with the behaviour of other components. This change brings it into compliance.


cgmb:
84e35103b7389bffabfe1757081c7a444872cca6 was merged to develop in #208.

PR #204, opened 2024-06-18
Fix names for header-only components when built statically

lawruble13:
Statically built packages may have different dependencies than the non-static packages, even for header only components whose contents are otherwise identical. As such, the generated packages should have separate names.


cgmb:
5a862812cf5625b1c9f2c90c6b0dc28514902bec was merged to develop in #208.

PR #212, opened 2024-07-19
Don't set `CPACK_RPM_MAIN_COMPONENT' for ASAN builds

lawruble13:
Setting `CPACK_RPM_MAIN_COMPONENT` to any component causes the package name for that component to be automatically generated from `CPACK_PACKAGE_NAME` (ignoring the other work we've done to correct the name), which in particular causes the runtime packages for ASAN builds to not be generated with the ASAN suffix.

In order to minimize the impact, I'm disabling this call only in the case that we're doing an ASAN build, I don't want to break any other spots where it is actually working correctly, but I think we should investigate removing this entirely.


cgmb (commented):
Is the static build going to have the same issue, as it also has a suffix?

It appears that the docs are failing to build on Ubuntu 20.04, but this appears to be a pre-existing issue, as it occurs on the branch as well.

[Note that the CMake behaviour was introduced in CMake 3.8](https://gitlab.kitware.com/cmake/cmake/-/commit/758f58b4a18fbb512edd1baf520017e816fee20b), and we use a newer version of cmake than that on all platforms so we can expect this behaviour to be consistent.

cgmb (approved):
> Is the static build going to have the same issue, as it also has a suffix?

Oh. Of course not, as there is no runtime component. LGTM.

lawruble13:
> Oh. Of course not, as there is no runtime component. LGTM.

Was typing pretty much exactly this, beat me to it 😄 

I was less concerned with inconsistent behaviour across CMake versions, and more across different builds. I think everything we've got set up should ensure the runtime package ends up always being named without a suffix, but I couldn't test that to verify at the time and we're short on time to have our fix cause something else to break.

pramenku:
Also @lawruble13  Please get this PR merge to Mainline using regular promotion process ASAP. Thanks,

PR #213, opened 2024-07-22
Restore previous rocm_install_targets behavior on Windows

apwojcik:
On Windows, we need to behave the old way when installing build artifacts—binaries, CMake configs, and development files.


ahsan-ca (approved):

cgmb (approved):
I approve on the condition that you explain what it is about the current behaviour that causes problems on Windows.

apwojcik:
> I approve on the condition that you explain what it is about the current behaviour that causes problems on Windows.

The problem is in the fixed DLL search order. Please read more here: https://learn.microsoft.com/en-us/windows/win32/dlls/dynamic-link-library-search-order
On Windows, all DLLs must be in an application's runtime directory unless loaded at runtime by the full path. 

Each dynamically linked DLL is subject to the DLL search order procedure. There is no RPATH on Windows.

eidenyoshida:
@apwojcik is this change still required? If so can you update the branch and re-assess for merge. Thanks

apwojcik:
> @apwojcik is this change still required? If so can you update the branch and re-assess for merge. Thanks

Yes, this is required. I thought the PR was already merged!

PR #237, opened 2024-10-24
Add deprecation message to ROCMConfig

lawruble13:
Add a deprecation message for use of `find_package(ROCM)`


eidenyoshida:
@lawruble13  can you have a look at the linting issues

cgmb:
It looks good, but we may need to update the test.

test/fail/rename-compatibility.cmake
```diff
@@ -0,0 +1,9 @@
+# ######################################################################################################################
+# Copyright (C) 2023 Advanced Micro Devices, Inc.
```
cgmb: Should this be 2024?

cgmb (approved):

test/fail/rename-compatibility.cmake
lawruble13: Yes, it should, updated.

PR #241, opened 2024-11-22
Update upper end of cmake_minimum_required range

lawruble13:
CMake 3.31 is dropping support for cmake_minimum_required for versions less than 3.10, but adding 3.25 (which is tested on math-ci) indicates that we test compatibility with 3.25, which causes those warnings to be suppressed.


lawruble13:
@pfultz2 I updated CMP0079 to NEW, as it throws a deprecation error when adding the upper end of the range. Is there a reason that needed to have the old behaviour?

pfultz2:
> Is there a reason that needed to have the old behaviour?

Either behavior works for ROCMTest module, however, I dont know if some downstream component is relying on the old behavior. If it doesnt break anyone, then using NEW behaviour is a better choice.

.github/workflows/test.yml
```diff
@@ -36,6 +36,7 @@ jobs:
           - macos-12
           - windows-2019
         cmake-version:
+          - 3.31.0
           - 3.20.0
           - 3.10.2
```
pfultz2: We should just set this to 3.16.3, which is the version used in ubuntu 20.04, then we could just update the minimum version to 3.16 in the test files.

.github/workflows/test.yml
```diff
@@ -36,6 +36,7 @@ jobs:
           - macos-12
           - windows-2019
         cmake-version:
+          - 3.31.0
           - 3.20.0
```
pfultz2: This should also probably be set to 3.22.1 which is the version in ubuntu 22.04.

share/rocmcmakebuildtools/cmake/ROCMTest.cmake
```diff
@@ -31,7 +31,7 @@ macro(rocm_enable_test_package NAME)
 endmacro()
 
 if(POLICY CMP0079)
-    cmake_policy(SET CMP0079 OLD)
+    cmake_policy(SET CMP0079 NEW)
 endif()
```
cgmb, line 35: Isn't it new by default if it exists?

cgmb (approved):

share/rocmcmakebuildtools/cmake/ROCMTest.cmake
lawruble13: `cmake_minimum_required` sets this to a value matching the behaviour from that version, so the new behaviour needs to be set to work with newer versions.

PR #257, opened 2025-01-09
Add rocm-core dependency to all generated packages, not just the base one

lawruble13:
To keep with the package standardization, all ROCm packages should have a direct dependency on rocm-core (transitive dependencies are insufficient). This PR causes the rocm-core package to be added as a dependency on all components registered via `rocm_package_setup_component` or `rocm_package_setup_client_component`.


cgmb (approved):
LGTM

PR #262, opened 2025-02-21
SWDEV-516659 - Ensure the list of recommended and suggested packages are separated by comma

raramakr:
A list in cmake is a semicolon(;) separated group of strings. Using semicolon in the cpack recommends and suggests variables will result in package installation failures. Use comma separated list


lawruble13:
Why is this required, and why should it be in ROCmCMakeBuildTools? We correctly handle these variables when we're dealing with them, and I don't think it's our place to be correcting variables set by the user even if they're not set correctly.

raramakr:
Will abandon this  PR and fix in the respective component Cmakelist.txt

PR #271, opened 2025-06-19
Update Ubuntu runner to 22.04

lawruble13:


cgmb (changes requested):
The test stage is still marked as running on ubuntu-20.04

PR #274, opened 2025-07-29
License File Folder Name Update to Package Name

arvindcheru:
Proposed Changes:
- Update License File folder update to Package Name

### DIFF
--- a/share/rocmcmakebuildtools/cmake/ROCMCreatePackage.cmake
+++ b/share/rocmcmakebuildtools/cmake/ROCMCreatePackage.cmake
@@ -473,6 +473,9 @@ macro(rocm_setup_license HEADER_ONLY)
                 DESTINATION share/doc/${_rocm_cpack_package_name}-asan
             )
         elseif((ROCM_USE_DEV_COMPONENT AND ${HEADER_ONLY}) OR NOT BUILD_SHARED_LIBS)
+	    if("${HDRONLY_LICENSE_INSTALLDIR}" STREQUAL "")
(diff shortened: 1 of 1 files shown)


share/rocmcmakebuildtools/cmake/ROCMCreatePackage.cmake
```diff
@@ -465,17 +467,20 @@ macro(rocm_setup_license HEADER_ONLY)
...
             install(
                 FILES ${CPACK_RESOURCE_FILE_LICENSE}
                 DESTINATION share/doc/${_rocm_cpack_package_name}-asan
             )
         elseif((ROCM_USE_DEV_COMPONENT AND ${HEADER_ONLY}) OR NOT BUILD_SHARED_LIBS)
+            message(STATUS "rocm-cmake: set install license file to ${HDRONLY_LICENSE_INSTALLDIR}.")
             install(
                 FILES ${CPACK_RESOURCE_FILE_LICENSE}
-                DESTINATION share/doc/${_rocm_cpack_package_name}
+                DESTINATION share/doc/${HDRONLY_LICENSE_INSTALLDIR}
```
cgmb, line 482: Just to be safe, can we keep the old behaviour when `HDRONLY_LICENSE_INSTALLDIR` is unset?

arvindcheru: Thank you for the comment, I tried to address in the latest patch. 
There are few failures I am facing in my testing, (even without this change) somehow the folder name generated is not consistently working for all package, ex: in case multiple packages from same module, it is random taking one of package name. Trying to check this now

PR #275, opened 2025-08-15
Add --hip-link to target_id check

lamb-j:
The new driver in clang requires linking against the HIP libraries, so we need to add a --hip-link option here to enforce linking

Without it, we'll get an error on the link step of check_cxx_compiler_flag() when calling with "-xhip" and "--offload-arch"

Note: The link step isn't really needed for this use case, as invalid targets will fail the compilation step. But check_cxx_compiler_flag() performs both compiling and linking by default


cgmb (approved):

jhuber6 (approved):
I think this is reasonable to require that the link step has the hip libraries. The reason the change to the new driver broke this behavior is because it unconditionally emits HIP runtime calls even with no kernel calls, the previous version would skip those if they weren't present.

jhuber6 (approved):

PR #281, opened 2026-01-15
Allow suppression of warnings in ROCMChecks

bstefanuk:
## Motivation

Warnings caused by third-party dependencies that are out of the control of a project cannot be suppressed and clutter the build output. 

## Technical Details

Only adds the warning statement if either ROCM_WARN_TOOLCHAIN_VAR or ROCM_ERROR_TOOLCHAIN_VAR are set, which enables nesting dependency fetches inside of scoped `block` calls to suppress warnings found by ROCMChecks.

## Test Plan

Standard CI testing. Manual function testing for hipblaslt confirms this pattern's validity.

## Test Result

Warnings are properly suppressed when ROCM_WARN_TOOLCHAIN_VAR is set to OFF.

## Submission Checklist

- [x] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.


lawruble13 (approved):
This looks good to me, for future work we have had some discussions about preferring to define methods which do explicitly scoped suppression of these warnings (for initial work on something like this refer to https://github.com/ROCm/rocm-cmake/pull/156).

PR #285, opened 2026-04-09
Add versioning for Windows

ivarusic-amd:
## Motivation

Adding versioninfo for migraphx binaries on Windows

## Technical Details

As discussed here https://github.com/ROCm/AMDMIGraphX/pull/4632; added versioning to rocm-cmake.


share/rocmcmakebuildtools/cmake/ROCMVersionResource.cmake
```diff
@@ -0,0 +1,179 @@
...
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#####################################################################################
```
pfultz2: That is not the way we add the license header.

share/rocmcmakebuildtools/cmake/ROCMVersionResource.cmake
```diff
@@ -0,0 +1,179 @@
...
+    [COPYRIGHT     <string>]   # defaults to "Copyright (c) 2015-<year> Advanced Micro Devices, Inc. All rights reserved."
+    [FILENAME      <string>]   # defaults to the target output name + .dll/.exe
+  )
+
+Example:
+  rocm_add_version_resource(
+    TARGET      mylib
+    DESCRIPTION "My ROCm Library"
+  )
+#]=======================================================================]
```
pfultz2: Documentation goes under docs/src

share/rocmcmakebuildtools/cmake/ROCMVersionResource.cmake
```diff
@@ -0,0 +1,179 @@
...
+
+function(rocm_add_version_resource)
```
pfultz2: This should go into ROCMSetupVersion.cmake.

pfultz2 (changes requested):
Tests need to be added and docs updated.

apwojcik:
> Tests need to be added and docs updated.

For the test, we need the MSVC resource compiler, which requires installing the MSBuild tools in Docker.

share/rocmcmakebuildtools/cmake/ROCMSetupVersion.cmake
```diff
@@ -1,6 +1,11 @@
 # ######################################################################################################################
-# Copyright (C) 2017 Advanced Micro Devices, Inc.
+# Copyright (C) 2026 Advanced Micro Devices, Inc.
 # ######################################################################################################################
+if(WIN32)
+    cmake_minimum_required(VERSION 3.17 FATAL_ERROR)
+endif()
```
pfultz2: Dont add this to the cmake files.

pfultz2:
> > Tests need to be added and docs updated.
> 
> For the test, we need the MSVC resource compiler, which requires installing the MSBuild tools

I think that is already installed. It has MSBuild version 17.14.36510.44 as listed [here](https://github.com/actions/runner-images/blob/main/images/windows/Windows2022-Readme.md).

> in Docker.

We dont use docker in the CI, although it does look like its available.

apwojcik:
> > > Tests need to be added and docs updated.
> > 
> > 
> > For the test, we need the MSVC resource compiler, which requires installing the MSBuild tools
> 
> I think that is already installed. It has MSBuild version 17.14.36510.44 as listed [here](https://github.com/actions/runner-images/blob/main/images/windows/Windows2022-Readme.md).
> 
> > in Docker.
> 
> We dont use docker in the CI, although it does look like its available.

I looked into testing infrastructure, but I don't know how to add the test you requested.

ivarusic-amd:
> > > Tests need to be added and docs updated.
> > 
> > 
> > For the test, we need the MSVC resource compiler, which requires installing the MSBuild tools
> 
> I think that is already installed. It has MSBuild version 17.14.36510.44 as listed [here](https://github.com/actions/runner-images/blob/main/images/windows/Windows2022-Readme.md).
> 
> > in Docker.
> 
> We dont use docker in the CI, although it does look like its available.



> > > Tests need to be added and docs updated.
> > 
> > 
> > For the test, we need the MSVC resource compiler, which requires installing the MSBuild tools
> 
> I think that is already installed. It has MSBuild version 17.14.36510.44 as listed [here](https://github.com/actions/runner-images/blob/main/images/windows/Windows2022-Readme.md).
> 
> > in Docker.
> 
> We dont use docker in the CI, although it does look like its available.

Tests added for both dll/exe

pfultz2 (approved):

PR #286, opened 2026-05-12
Dont set config file by default and add TIDY_ARGS to set additional flags

pfultz2:
## Motivation

Setting the config file overwrites the default search for .clang-tidy files. It should still find the top-level .clang-tidy file but this will allow overriding it in certain directories.

## Submission Checklist


share/rocmcmakebuildtools/cmake/ROCMClangTidy.cmake
```diff
@@ -127,6 +127,10 @@ macro(rocm_enable_clang_tidy)
             list(APPEND CLANG_TIDY_EXTRA_ARGS -extra-arg=-Xclang "-extra-arg=${ARG}")
         endforeach()
 
+        foreach(ARG ${PARSE_TIDY_ARGS})
+            list(APPEND CLANG_TIDY_EXTRA_ARGS "${ARG}")
+        endforeach()
+
```
amd-isparry, line 133: Why 3 lines? This uses multiple calls to list(APPEND), which is itself an order n operation, so this is O(n^2) because of the memory allocations.

Just use 
 
    list(APPEND CLANG_TIDY_EXTRA_ARGS ${PARSE_TIDY_ARGS})

which will do the same thing, in particular remove empty list elements.
AI generated gist at https://gist.github.com/amd-isparry/7af0ee4f640e8d5d4fd8b97282c18831

PR #288, opened 2026-06-24
Install pdb files

pfultz2:
## Submission Checklist

### DIFF
--- a/share/rocmcmakebuildtools/cmake/ROCMInstallTargets.cmake
+++ b/share/rocmcmakebuildtools/cmake/ROCMInstallTargets.cmake
@@ -233,7 +233,8 @@ function(rocm_install_targets)
                 DESTINATION ${LIB_INSTALL_DIR}
                 COMPONENT ${development}
         )
-        if(WIN32 AND
+        if((CMAKE_CXX_COMPILER_FRONTEND_VARIANT STREQUAL "MSVC" OR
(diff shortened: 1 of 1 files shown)


ivarusic-amd:
@pfultz2 I think, Windows CI fails because $<TARGET_PDB_FILE:...> is only supported by MSVC-compatible toolchains
  Two CI configurations hit this error:

  1. Ninja + clang++
  2. Ninja + default - on the GitHub runner, Ninja picks up MinGW/GCC from PATH 
 Both of those **does not** produce PDB files (at least will not work with <TARGET_PDB_FILE:...>) .

  To fix this we need an MSVC guard around the TARGET_PDB_FILE usage, so PDB installation is skipped for compilers that
  can't produce them. 

  Additionally, we can update the CI to use MSVC-compatible compilers for all Windows jobs: switch clang++ to clang-cl
  and add msvc-dev-cmd so Ninja finds MSVC instead of MinGW. This way all Windows CI configurations actually test PDB
  installation rather than just skipping it.

ivarusic-amd:
I've tested this in draft PR here: https://github.com/ROCm/rocm-cmake/pull/290

ivarusic-amd:
MSVC flag is not enough;        
```
 if(WIN32 AND
            (CMAKE_CXX_COMPILER_FRONTEND_VARIANT STREQUAL "MSVC" OR
             CMAKE_C_COMPILER_FRONTEND_VARIANT STREQUAL "MSVC") AND
```

